PR #2, opened 2023-04-18
initial draft (to be reviewed) for API def.

kmrdhruv:
Initial API draft for tenant/topic/subscription entity mgmt, produce and dlq consume.


docs/api.yaml
```diff
@@ -0,0 +1,925 @@
...
+    capacity_policy:
+      type: object
+      properties:
+        max_throughput:
+          type: integer
+          description: "Maximum throughput in Mbps."
+          example: 10
+        max_storage:
+          type: integer
+          description: "Maximum size in MB"
```
gauravAshok, line 519: if we have avg msg size & throughput & retention policy. Will this info be useful? It looks to be inferable.

docs/api.yaml
```diff
@@ -0,0 +1,925 @@
...
+          type: integer
+          example: 24
+          default: 48
+          maximum: 76
+          
+    capacity_policy:
+      type: object
+      properties:
+        max_throughput:
+          type: integer
```
gauravAshok, line 514: should we make it more human friendly, like 
10Kb, 1024B, 1MB. The units can be assumed to be per sec. and we only ask for string that can be parsed as unit of some bytes.

or we reduce the scale?. just ask KB or B. MB might be too large.

docs/api.yaml
```diff
@@ -0,0 +1,925 @@
...
+          description: "maxmimum retry interval in ms."
+          example: 1500
+        multiplier:
+          type: integer
+          description: "step for iteration"
+          example: 10
+        retry_attempts:
+          type: integer
+          description: "maximum retry attempts allowed."
+          maximum: 3
```
gauravAshok, line 555: so there are "small" / frequent retries that we we can do inline (i.e. during consumption itself), but there are longer retries that we can do using topics.

How do configs map to the above behaviour?

gauravAshok: As discussed, multiplier is just that, its description may not be be correct. 
So, now the current policy will be backed by topics. 
From implementation point of view, there may be scope of doing some initial retries locally only, before we park such messages in to RQ. do we need to make that explicit as part of policy.

docs/api.yaml
```diff
@@ -0,0 +1,925 @@
...
+      properties:
+        max_parallelism:
+          type: integer
+          example: 10
+        recovery_allocation:
+          type: number
+          example: 0.5
+        dlq_from_recovery:
+          type: number
+          example: 0.5
```
gauravAshok, line 577: I am not too inclined on "dlq_from_recovery" ratio. Make things slightly complicated, too fine grained which users probably can't manage.

since messages in RQ & DLQ are "failed" messages, users might want to prioritize one over the other, dependeing on their usecase. Cant we just say, "recovery_allocation" generally applies to recovery of failed messages, doen't matter RQ / DLQ. By default RQ messages will be prioritized. 
If need be we can just provide the toggle to prioritize RQ or DLQ over the other, (as this probably will be business usecase specific). 

Also, just to point out the behaviour, one nuance that a plain "0.5" ratio doesn't capture, is that it is max allocation. if there are not enough RQ / DLQ messages, the unused allocation is just kind of used for main messages. Similar behaviour when prioritizing RQ over DLQ, if there are not enough messages/parallelism in RQ, then DLQ will be consumed from pending parallelism quota.

docs/api.yaml
```diff
@@ -0,0 +1,925 @@
...
+        max_error_threshold:
```
gauravAshok, line 578: In case of grouped subscription, will the error_rate be calculated in terms of failed groups, instead of messages failing to RQ / DLQ. right?

docs/api.yaml
```diff
@@ -0,0 +1,925 @@
...
+          $ref: '#/components/schemas/endpoint'
+        retention_policy:
+          $ref: '#/components/schemas/retention_policy'
+        retry_policy:
+          $ref: '#/components/schemas/retry_policy'
+        dlq_policy:
+          $ref: '#/components/schemas/dlq_policy'
+        consumption_policy:
+          $ref: '#/components/schemas/consumption_policy'
+        version:
```
gauravAshok, line 689: when we say version, we mean schema version right?

docs/api.yaml
```diff
@@ -0,0 +1,925 @@
...
+          description: "Identity of application which produced this message. It will be auto populated by the platform."
+        message_id:
+          type: string
+          format: uuid
+        deduplication_id:
+          type: string
+          description: "Not used for now. Intended for message deduplication semantics."
+        properties:
+          $ref: '#/components/schemas/property_bag'
+        groupkey:
```
gauravAshok, line 712:
```suggestion
        group_key:
```

docs/api.yaml
```diff
@@ -0,0 +1,925 @@
...
+            schema:
+              $ref: '#/components/schemas/state_request'
+      responses:
+        '200':
+          description: OK
+        default:
+          $ref: '#/components/responses/error_response'
+  
+          
+  /v1/tenants/{tenant_name}/subscriptions/{sub_name}/status:
```
gauravAshok, line 411: how is this status different from state.

This state vs status, led me to search the difference between them. I think I align with this https://softwareengineering.stackexchange.com/questions/219351/state-or-status-when-should-a-variable-name-contain-the-word-state-and-w

I can understand the active, inactive (/ blocked), running, stopped. being the status of the subscription. Similarly, active, inactive (/blocked) for the topic status.

What is consumer_status exactly. Is this for internal details of the consumer, lag and other stuff? Internal Stats?

docs/api.yaml
```diff
@@ -0,0 +1,925 @@
...
+
+  /v1/tenants/{tenant_name}/subscriptions/{sub_name}/dlt:
+    parameters:
+      - $ref: '#/components/parameters/tenanat_param'
+      - $ref: '#/components/parameters/sub_param'
+      - $ref: '#/components/parameters/http_code'
+      - $ref: '#/components/parameters/from_param'
+      - $ref: '#/components/parameters/to_param'
+      - $ref: '#/components/parameters/msg_id_param'
+      - $ref: '#/components/parameters/group_id_param'
```
gauravAshok, line 441: group_id vs group_key. we can stick to one.

docs/api.yaml
kmrdhruv: intresting.. I was thinking of it as "state" being more like state-machine state, status being more detailed running info. kind of single value and multi-value attribute.
Both info as of now intended for external consumption.

docs/api.yaml
kmrdhruv: aligned .. any preferences ?

docs/api.yaml
```diff
@@ -0,0 +1,925 @@
...
+              schema:
+                type: array
+                items:
+                  $ref: '#/components/schemas/recieved_message'
+        default:
+          $ref: '#/components/responses/error_response'
+          
+    put:
+      description: >-
+        Unsideline messages from DLT for further consumption.
```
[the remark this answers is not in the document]
kmrdhruv: As path includes "dlt", not applicable for MQ/RQ.
get is returning msgs,
put is modifying the dlt internal state i.e. moving the pointer.

docs/api.yaml
kmrdhruv: i was more inclined towards max_storage than msg size as msg size might vary and tough to guess avg size. Also max storage might help in force purge or billing aspect.

docs/api.yaml
```diff
@@ -0,0 +1,925 @@
...
+          type: integer
+          description: "Maximum throughput in Mbps."
+          example: 10
+        max_storage:
+          type: integer
+          description: "Maximum size in MB"
+          example: 1500
+        message_size:
+          type: integer
+          description: "Maximum message size in KBs."
```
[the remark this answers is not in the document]
kmrdhruv: this i just added for reference/discussion. We should keep either of them.

docs/api.yaml
kmrdhruv: This was object versioning and not schema versioning.
Shall we cover schema version using API versioning ?

docs/api.yaml
```diff
@@ -0,0 +1,925 @@
...
+          properties:
+            factor:
+              type: number
+              description: "Factor by which qps needs to be increased/decreased."
+              example: 0.1
+            wait_seconds:
+              type: integer
+              description: "wait in seconds before attempting next increase/decrease step."
+              example: 10
+            ping_seconds:
```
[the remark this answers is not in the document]
kmrdhruv: may be needs renaming...
post complete throttle, consumer needs to continuously check at regular interval if service has come back and start recovery process. This regular interval is being called as ping_seconds.
Think of it as attempt to deliver the message continuously where failure doesn't move the msg to different queue.

docs/api.yaml
kmrdhruv: I was thinking of later.. as primarily MTTR and also rq/dq size will be determined by number of messages.

docs/api.yaml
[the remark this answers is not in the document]
kmrdhruv: Request tracking mechanism needs to be figured out.

docs/api.yaml
kmrdhruv: merge both retention and capacity, also only throughput.

PR #10, opened 2023-06-15
move rest API to vertx worker pool and docker environment for e2e test execution.

kmrdhruv:


setup/docker/compose.yml
```diff
@@ -0,0 +1,80 @@
...
+volumes:
+  pulsarconf:
+  pulsardata:
+  zkdata:
+  zklogs:
+  zkdatalog:
+
+services:
+  web:
+    image: vardhi.docker.registry/varadhi:latest
```
gauravAshok, line 17:
```suggestion
    image: varadhi.docker.registry/varadhi:latest
```
typo

setup/docker/compose.yml
```diff
@@ -0,0 +1,80 @@
...
+      driver: default
+      config:
+        - subnet: 10.5.5.0/24
+
+volumes:
+  pulsarconf:
+  pulsardata:
+  zkdata:
+  zklogs:
+  zkdatalog:
```
gauravAshok, line 13: anything expected here?

setup/docker/Dockerfile
```diff
@@ -0,0 +1,31 @@
+# base image for this should be jre based on instead of jdk.
```
gauravAshok, line 1: https://youtu.be/irV3KVIcoN0?t=223

setup/docker/compose.yml
```diff
@@ -0,0 +1,80 @@
...
+    restart: "no"
+    depends_on:
+      pulsar:
+        condition: service_healthy
+      zookeeper:
+        condition: service_healthy
+
+  pulsar:
+    image: apachepulsar/pulsar:3.0.0
+    profiles: [ "test", "dev", "pulsar" ]
```
gauravAshok, line 33: what is these profiles used for?

setup/docker/Dockerfile
```diff
@@ -0,0 +1,31 @@
...
+RUN sed -i  "s|'.*zkConfig.yml'|'/etc/varadhi/zkConfig.yml'|" /etc/varadhi/configuration.yml
+# RUN sed -i  "s|127.0.0.1|10.5.5.3|" /etc/varadhi/pulsarConfig.yml
+# RUN sed -i  "s|127.0.0.1|10.5.5.4|" /etc/varadhi/zkConfig.yml
+RUN  if [ "$ENV" = "test" ] ; then sed -i  "s|127.0.0.1:8081|10.5.5.3:8080|" /etc/varadhi/pulsarConfig.yml ; fi
+RUN  if [ "$ENV" = "test" ] ; then sed -i  "s|127.0.0.1|10.5.5.4|" /etc/varadhi/zkConfig.yml ; fi
+
+WORKDIR /usr/share/varadhi
+
+# Start Varadhi server
+CMD java -cp ./*:dependencies/* com.flipkart.varadhi.Server /etc/varadhi/configuration.yml
```
gauravAshok, line 31: do we need to create a user for running the application?

server/src/main/resources/configuration.yml
```diff
@@ -3,8 +3,10 @@ vertxOptions:
   workerPoolSize: 1
   internalBlockingPoolSize: 1
 
-deploymentOptions:
-  instances: 1
+restVerticleDeploymentOptions:
+  instances: 2
```
gauravAshok, line 7: in case of worker pool based, what does having multiple instances mean?

setup/docker/compose.yml
kmrdhruv: for data persistence across compose up/down.

setup/docker/compose.yml
kmrdhruv: grouping, containers under that profile would be brought up as part of compose up/down. (docker compose --profile)

setup/docker/Dockerfile
kmrdhruv: likely yes, so that it doesn't run under default.
I will take it up.

server/src/main/resources/configuration.yml
kmrdhruv: in worker pool, instances are no concurrent in terms of execution.

setup/docker/Dockerfile
kmrdhruv: think of it as a test image... so didn't start with slim for both os and java to have all standard utilities by default for sometime. But yes, it needs to be brought down once we have stable working setup.

server/src/main/java/com/flipkart/varadhi/utils/ResponseExtension.java
```diff
@@ -9,6 +11,8 @@
 public class ResponseExtension {
```
gauravAshok, line 11: This is technically a RoutingContext's Extension :)

gauravAshok, line 11: This is technically a RoutingContext's Extension :)

PR #12, opened 2023-06-20
Introducing ZK Persistence Methods Delete & List

### LINKED ISSUE #14: Implement Create/List/Get/Delete generic APIs at ZK persistent layer.

yagami-kunal:


server/src/main/java/com/flipkart/varadhi/db/ZKPersistence.java
```diff
@@ -73,6 +76,27 @@ public boolean exists(String resourceKey) {
...
+        } catch (Exception e) {
+            log.error(String.format("Failed to delete entity %s ", resourcePath), e);
+            throw new VaradhiException(e);
+        }
+    }
+
+    public List<String> list(String resourceKey) {
+        String resourcePath = getResourcePath(resourceKey);
+        try {
+            if (!exists(resourceKey)) return Collections.emptyList();
```
gauravAshok, line 92: is this alright to return empty here? Shouldn't we throw error? Considering that caller wanted to know the children under a path, but the path does not even exist.

gauravAshok (changes requested):

server/src/main/java/com/flipkart/varadhi/db/ZKPersistence.java
```diff
@@ -73,6 +76,27 @@ public boolean exists(String resourceKey) {
...
+            return zkCurator.getChildren().forPath(resourcePath);
+        } catch (Exception e) {
+            log.error(String.format("Failed to list entity on %s ", resourcePath), e);
+            throw new VaradhiException(e);
+        }
+    }
+
     private String getResourcePath(String resourceKey) {
         return String.format("%s%s", BASE_PATH, resourceKey);
```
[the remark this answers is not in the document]
yagami-kunal: yes, but the topic resource path starts with "/" so this suffices `uniqueKeyPath` in topic resource for reference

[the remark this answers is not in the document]
gauravAshok: We shouldn't think about the rest layer which is a higher layer. ZkPersistence being a core / low level layer, should be the one enforcing "some" contract on its methods and its params.

So, "resourceKey" should have some restrictions/rules, and the method probably needs to make some validations around that.

entities/src/main/java/com/flipkart/varadhi/entities/TopicResource.java
```diff
@@ -31,7 +31,7 @@ public TopicResource(
 
     @Override
     public String uniqueKeyPath() {
-        return String.format("/%s/%s/%s", RESOURCE_TYPE_NAME, project, getName());
+        return String.format("%s/%s/%s", RESOURCE_TYPE_NAME, project, getName());
```
kmrdhruv, line 34: @yagami-kunal e2e test is little fragile as of now. You can change the test to use uniqueKeyPath() in its duplicate topic assertion.

core/src/main/java/com/flipkart/varadhi/db/MetaStore.java
```diff
@@ -17,4 +19,9 @@ public interface MetaStore {
 
     VaradhiTopic getVaradhiTopic(String topicName);
 
+    void deleteVaradhiTopic(String topicName);
+
+    void deleteTopicResource(String projectName, String resourceName);
+
+    List<String> listVaradhiTopics(String projectName);
```
kmrdhruv, line 26: do not implement this API name as of now. Needs some more thought on listing of Varadhi Topic objects as project name is embedded with in topic name and it would require some parsing.

server/src/main/java/com/flipkart/varadhi/db/ZKMetaStore.java
```diff
@@ -27,7 +29,8 @@ public <T> String create(T resource, int version, String resourcePath) {
             log.debug("Persisted Entity({}) in ZK: {}", resourcePath, response);
             return response;
         } catch (Exception e) {
-            throw new VaradhiException(e);
+            log.error(String.format("Failed to create entity %s.", resource), e);
+            throw new MetaStoreException(e);
```
kmrdhruv, line 33: Add error message to MetaStoreException and remove log.error() -- this is the updated convention i.e. log only if being handled otherwise throw with required info.

This will be applicable to other methods as well.

server/src/main/java/com/flipkart/varadhi/db/ZKPathUtils.java
```diff
@@ -10,6 +10,10 @@ public static String getVaradhiTopicPath(String topicName) {
         return constructPath(VARADHI_TOPIC_NAME, topicName);
     }
 
+    public static String getVaradhiProjectPath(String projectName) {
+        return constructPath(TOPIC_RESOURCE_NAME, projectName);
+    }
+
     public static String getTopicResourceFQDN(String projectName, String topicName) {
         return String.join("/", projectName, topicName);
```
kmrdhruv, line 18: There is a bug in the code I checked in. This should not be path separator in above. But a name separator.  We can define a constant and use it accordingly. 

Since DB package is missing UTs, can we add some tests to start with as well.

kmrdhruv (changes requested):

server/src/test/java/com/flipkart/varadhi/db/ZKMetaStoreTests.java
```diff
@@ -0,0 +1,45 @@
...
+        zkMetaStore = new ZKMetaStore(zkCuratorFramework);
+    }
+
+    @AfterAll
+    public static void closeTest() throws IOException {
+        zkCuratorTestingServer.close();
+    }
+
+
+    @Test
```
kmrdhruv, line 38: can you add a test for delete and list as well ?

kmrdhruv (approved):

kmrdhruv (approved):

PR #13, opened 2023-06-21
Refactor CoreServices

kmrdhruv:


server/src/main/java/com/flipkart/varadhi/RestVerticle.java
```diff
@@ -10,29 +13,44 @@
...
+
         log.info("HttpServer Starting.");
         Router router = Router.router(vertx);
 
         FailureHandler failureHandler = new FailureHandler();
-        for (RouteDefinition def : coreServices.getRouteDefinitions()) {
+        for (RouteDefinition def : apiRoutes) {
             Route route = router.route().method(def.method()).path(def.path());
-            def.behaviours().forEach(d -> d.Configure(route, def, coreServices));
+            def.behaviours().forEach(behaviour -> {
```
gauravAshok, line 42: This may not be safe right? I think the handlers need to be configured in fixed order.

For eg: BodyHandler should not come before Auth handlers.

PR #27, opened 2023-06-27
Add python script for zookeeper benchmark

aayustark007-fk:
This script connects to zookeeper ensemble, loads data under a parent node and then measures the response time of get_children() call.

Various parameters are configurable and can be fetched by running `python zk_benchmark.py -h`
To speed up data loading, there are two options, multi-threaded dataloader and multi-process dataloader. Multi-process dataloader has some bugs so its not enabled by default.

Requirements: python 3.10+, tqdm, kazoo


scripts/zk_benchmark.py
```diff
@@ -0,0 +1,290 @@
...
+            futures_to_id = {exec.submit(self._process, parent, config): id for (id, config) in enumerate(configs)}
+            for future in concurrent.futures.as_completed(futures_to_id):
+                id = futures_to_id[future]
+                try:
+                    res = future.result()
+                except Exception as exc:
+                    print("process: {} generated an exception: {}".format(id, exc))
+
+    def _process(self, parent: str, child_node_config: ChildNodeConfig):
+        with DataLoader(self.config, zk_config=self.zk_config) as loader:
```
kmrdhruv, line 209: is this creating threadpool in processpool, expected ?

scripts/zk_benchmark.py
```diff
@@ -0,0 +1,290 @@
...
+        self.config = config
+        self.zk_config = zk_config
+    
+    def __enter__(self):
+        return self
+    
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        return True
+
+    def __get_chunked_configs(self, child_node_config: ChildNodeConfig) -> list[ChildNodeConfig]:
```
kmrdhruv, line 183: with default chunk_size of 10K, and target range of 10K, this will effectively generate single block ?
should default be smaller number say 1K ?

scripts/zk_benchmark.py
```diff
@@ -0,0 +1,290 @@
...
+    size = random.randrange(low, high)
+    return random_string(size)
+
+# Generator which generates the child node data based on params
+def node_generator(num: int, name="", min_name_len=-1, max_name_len=-1, data_size_bytes=1024) -> NodeGenerator:
+    gen = 0
+    while gen < num:
+        name_res = ""
+        if len(name) > 0:
+            name_res = "{}_{}".format(name, str(gen + 1))
```
kmrdhruv, line 29: this wouldn't take care of name length requirements, but that should be fine in case fixed name is needed. just add a comment for the same.

scripts/zk_benchmark.py
```diff
@@ -0,0 +1,290 @@
...
+            loader.load_data(run_config.name, run_config.child_node_config)
+
+        ## Measure step
+        if not skip_measure:
+            with ZkClientWrapper(config=self.zk_config, name="measure_zk") as client:
+                print("Num child nodes under path {}: {}".format(path, len(client.ls(path))))
+                with catchtime() as t:
+                    for _ in range(measure_samples):
+                        client.ls(path)
+                print("Latency get_children on path {}: {} ms".format(path, t.time / measure_samples))
```
kmrdhruv, line 244: in this case it wouldn't matter much, but it would be good to have it as min, max, avg  over the sample range.

kmrdhruv (changes requested):
move this to scripts/benchmarks directory and also add a small readme under same folder on how to install dependencies and run the benchmark as well.

scripts/zk_benchmark.py
```diff
@@ -0,0 +1,290 @@
...
+                data_bytes=child_node_config.data_bytes))
+            rem_size -= gen_size
+        
+        return result
+
+    def load_data(self, parent: str, child_node_config: ChildNodeConfig):
+        configs = self.__get_chunked_configs(child_node_config)
+        with concurrent.futures.ProcessPoolExecutor(max_workers=self.config.parallelism) as exec:
+            futures_to_id = {exec.submit(self._process, parent, config): id for (id, config) in enumerate(configs)}
+            for future in concurrent.futures.as_completed(futures_to_id):
```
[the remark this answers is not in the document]
aayustark007-fk: the API for multiprocess executor does not provide a way to pass timeout
trying to find a way to pass sigterm to subprocesses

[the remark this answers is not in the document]
kmrdhruv: may be kill process group pgid()

scripts/benchmark/README.md
```diff
@@ -0,0 +1,27 @@
+## Python scripts to benchmark the Metadata Stores
+
+### Currently, holds the script for benchmarking read performance of Zookeeper
+
+Requires: `Python==3.9.7+`
+
```
kmrdhruv, line 6: pip -r requirements.txt  ? needed ?

scripts/benchmark/zk_benchmark.py
```diff
@@ -312,6 +315,7 @@ def __run(self, run_config: BenchmarkRunConfig, skip_measure: bool, dataloading_
         
         ## Measure step
         self.__measure(path, measure_samples, skip_measure)
```
kmrdhruv, line 317: nit: __measure_list()

kmrdhruv (approved):
LGTM.
Do execute the benchmark on a standard zk setup and publish results.

scripts/zk_benchmark.py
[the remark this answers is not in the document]
aayustark007-fk: So, I read that by using with statement will cause executor.shutdown() to be called which will terminate the processes. My assumption is that if I send a SIGINT via ctrl+c it should cause program to terminate and the with block to get cleaned up.

But, while running I see that when zookeeper becomes unreachable, the multiprocess program behaves badly. The program gets stuck and upon SIGTERM the main process exits but some of the processes continue to exists as zombies. (sending logs to the terminal)

kmrdhruv (approved):

PR #45, opened 2023-09-03
Default AuthZ Provider Skeleton Code

aayustark007-fk:
- Implemented config driven roles and role binding
- Linking to the AuthorizationProvider interface and implemented isAuthorised method
- Added some basic roles and role bindings


server/src/main/java/com/flipkart/varadhi/auth/DefaultAuthorizationProvider.java
```diff
@@ -0,0 +1,92 @@
...
+     * Checks if the user {@code UserContext} is authorized to perform action {@code ResourceAction} on resource path.
+     *
+     * @param userContext information on the user, contains the user identifier
+     * @param action      the action being performed by the user which is to be authorized
+     * @param resource    the path of the resource on which the action is to be authorized
+     *
+     * @return {@code Future<Boolean>} a future result expressing True/False decision
+     */
+    @Override
+    public Future<Boolean> isAuthorized(UserContext userContext, ResourceAction action, String resource) {
```
kmrdhruv, line 34: Do we need this to be Async ?

server/src/main/java/com/flipkart/varadhi/auth/DefaultAuthorizationProvider.java
```diff
@@ -0,0 +1,92 @@
...
+        Map<ResourceType, String> resourceTypeToResourceId = getResourceIds(action, resource);
+        return Future.succeededFuture(resourceTypeToResourceId.entrySet().stream()
+                .anyMatch(entry -> isAuthorizedInternal(userContext.getSubject(), action, entry.getValue())));
+    }
+
+    private Map<ResourceType, String> getResourceIds(ResourceAction action, String resource) {
+        // TODO(aayush): need better way to do this?
+        Map<ResourceType, String> resultMap = new HashMap<>();
+        String[] segments = resource.split("/");
+        return resolve(ResourceType.ORG, action, List.of(segments), resultMap);
```
kmrdhruv, line 45: Since format is fixed and resolve also depends on the format, would it be better to do simple parsing instead of recursive call ?

server/src/main/resources/configuration.yml
```diff
@@ -23,12 +23,41 @@ authentication:
...
       ignoreExpiration: true
 
-authorizationEnabled: false
+authorizationEnabled: true
 
 authorization:
   superUsers: [ "thanos" ]
-  providerClassName:
-  providerOptions:
+  providerClassName: "com.flipkart.varadhi.auth.DefaultAuthorizationProvider"
```
kmrdhruv, line 30: fine for POC, but for prod it might be better to extract these to authz specific file and have a pointer in the main config file.

server/src/main/java/com/flipkart/varadhi/auth/DefaultAuthorizationProvider.java
```diff
@@ -0,0 +1,92 @@
...
+     * @param userContext information on the user, contains the user identifier
+     * @param action      the action being performed by the user which is to be authorized
+     * @param resource    the path of the resource on which the action is to be authorized
+     *
+     * @return {@code Future<Boolean>} a future result expressing True/False decision
+     */
+    @Override
+    public Future<Boolean> isAuthorized(UserContext userContext, ResourceAction action, String resource) {
+        // parse the resource path based on the action and return the final resourceIDs for each resourceType
+        Map<ResourceType, String> resourceTypeToResourceId = getResourceIds(action, resource);
```
kmrdhruv, line 36: would it be more beneficial to evaluate it from root to leaf or opposite, thoughts ?

server/src/main/java/com/flipkart/varadhi/auth/DefaultAuthorizationProvider.java
```diff
@@ -0,0 +1,92 @@
...
+                .getOrDefault(subject, new ArrayList<>());
+    }
+
+    private boolean isAuthorizedInternal(String subject, ResourceAction action, String resourceId) {
+        return getRolesForSubject(subject, resourceId).stream()
+                .anyMatch(role -> doesActionBelongToRole(role, action));
+    }
+
+    private boolean doesActionBelongToRole(String roleId, ResourceAction action) {
+        return configuration.getRoles().getOrDefault(roleId, new ArrayList<>()).contains(action);
```
kmrdhruv, line 90: for debugging purpose, it would be good maintain (log) the information which node and which role returned true for authorised call.

kmrdhruv (changes requested):

server/src/main/java/com/flipkart/varadhi/auth/DefaultAuthorizationProvider.java
aayustark007-fk: The current authorization provider interface returns Future.
I think thats so that our handler is not blocking. https://vertx.io/docs/vertx-core/java/#golden_rule

server/src/main/resources/configuration.yml
aayustark007-fk: sure will do that

server/src/main/java/com/flipkart/varadhi/auth/DefaultAuthorizationProvider.java
aayustark007-fk: Currently we iterate over map entries, so the evaluation order here is undefined.

One problem them I'm facing is that in case of project, how do I figure out the parent entity ID?

Some approaches I have in mind:

1. Pass only the project_id as resource (as being currently done in RequiredAuthorization) and the Authorization Provider has the required mapping to the parent entity. So, when we fetch the role_bindings for a given resource_id, we get the role_bindings at that resource_id level along with the role_bindings at the parent level till the root node.
2. Or, since this is a Default barebones impl, we just do away with the requirement of inheriting permissions and check only at the resource_ids being passed.

Now coming to the original question, what would be better? Root to Leaf or opposite?
I'd say, considering the points above, opposite evaluation will be much better as we are mostly being passed the last set of identifiers through which we will mostly move up the hierarchy. Evaluating Root to Leaf will involve the extra work of figuring out the entire hierarchy at once and then evaluating.

What are your thoughts?

server/src/main/java/com/flipkart/varadhi/auth/DefaultAuthorizationProvider.java
aayustark007-fk: done

server/src/main/java/com/flipkart/varadhi/auth/DefaultAuthorizationProvider.java
aayustark007-fk: agree, will make the changes based on the outcome of https://github.com/flipkart-incubator/varadhi/pull/45#discussion_r1335046085

server/src/main/java/com/flipkart/varadhi/auth/DefaultAuthorizationProvider.java
aayustark007-fk: Updated the code, we are evaluating in lazy manner from leaf to root node

entities/src/main/java/com/flipkart/varadhi/auth/ResourceType.java
```diff
@@ -1,6 +1,7 @@
 package com.flipkart.varadhi.auth;
 
 public enum ResourceType {
+    ROOT("root"),
```
aayustark007-fk, line 4: required for bootstrapping and org create

server/src/main/java/com/flipkart/varadhi/auth/DefaultAuthorizationProvider.java
```diff
@@ -0,0 +1,139 @@
...
+import org.apache.commons.lang3.tuple.Pair;
+
+
+@Slf4j
+public class DefaultAuthorizationProvider implements AuthorizationProvider {
+    private DefaultAuthorizationConfiguration configuration;
+    private volatile boolean initialised = false;
+
+    @Override
+    public Future<Boolean> init(AuthorizationOptions authorizationOptions) {
```
aayustark007-fk, line 22: Was planning to read the file as json object and pass to the providers. But, decided against it because I think it would be better to have control at the provider end in case they wish to specify the configuration as JSON instead of YAML or even interpret the configFile path as an URI and fetch the config over the internet or file server.

aayustark007-fk: with this we can also give configFile as:
`configFile: "https://raw.githubusercontent.com/acme-org/acme-project/customconfig.json"`

And the provider can then make an http call to fetch and parse the config.

kmrdhruv (changes requested):

entities/src/main/java/com/flipkart/varadhi/auth/ResourceType.java
aayustark007-fk: Idea is that when there is no Org defined, we still need a resource node to hold the bootstrap role bindings. Thats why a ROOT is defined which will hold role bindings required for org creation and role assignment. We can also drive super-users by adding role bindings to this node.

entities/src/main/java/com/flipkart/varadhi/auth/DefaultAuthorizationConfiguration.java
```diff
@@ -0,0 +1,12 @@
...
+
+import lombok.Data;
+
+import java.util.List;
+import java.util.Map;
+
+@Data
+public class DefaultAuthorizationConfiguration {
+    private Map<String, List<String>> roles;
+    private Map<String, Map<String, List<String>>> roleBindings;
```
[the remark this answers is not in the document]
aayustark007-fk: This is just a temporary configuration. Once we have proper CRUD APIs for roles and role bindings, we can remove this config. We can then only drive the ability to provide bootstrap role bindings via these configs.

server/src/main/java/com/flipkart/varadhi/auth/DefaultAuthorizationProvider.java
```diff
@@ -0,0 +1,139 @@
...
+        log.debug(
+                "Checking authorization for subject [{}] and action [{}] on resource [{}]", subject, action,
+                resourceId
+        );
+        return getRolesForSubject(subject, resourceId).stream()
+                .anyMatch(role -> doesActionBelongToRole(subject, role, action));
+    }
+
+    private List<String> getRolesForSubject(String subject, String resourceId) {
+        return configuration.getRoleBindings()
```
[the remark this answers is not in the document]
aayustark007-fk: Once CRUD APIs are available for role and role-binding, this part of the code will call the GetRoleBinding API instead.

server/src/main/java/com/flipkart/varadhi/auth/DefaultAuthorizationProvider.java
```diff
@@ -0,0 +1,139 @@
...
+     * @param resourcePath uri of the resource
+     * @return List of pairs having resource type to its id. List is used so that we can impose ordering from leaf to root nodes.
+     */
+    private List<Pair<ResourceType, String>> resolveOrderedFromLeaf(ResourceAction action, String resourcePath) {
+        String[] segments = resourcePath.split("/");
+
+        // build the list in reverse order specified: ROOT -> ORG -> TEAM -> PROJECT -> TOPIC|SUBSCRIPTION|QUEUE
+        List<Pair<ResourceType, String>> resourceIdTuples = new ArrayList<>();
+        // handle leaf node case
+        if (isActionOnLeafNode(action)) {
```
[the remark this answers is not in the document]
aayustark007-fk: Not a special case, we want to know if there are any topic or subscription id provided in the resource path. Eg: Say in case of project create, there will be no topic or subscription id present so no leaf node is added in the list of tuples.

server/src/main/java/com/flipkart/varadhi/auth/DefaultAuthorizationProvider.java
```diff
@@ -0,0 +1,139 @@
...
+            resourceIdTuples.add(Pair.of(action.getResourceType(), getLeaf(segments)));
+        }
+        resourceIdTuples.add(Pair.of(ResourceType.PROJECT, getProject(segments)));
```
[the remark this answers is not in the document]
aayustark007-fk: I think we can omit adding keys which are blank. Will make the change

server/src/main/resources/configuration.yml
```diff
@@ -27,8 +27,9 @@ authorizationEnabled: false
 
 authorization:
   superUsers: [ "thanos" ]
```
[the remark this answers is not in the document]
aayustark007-fk: my understanding was that for super-users, we will skip the authZ checks entirely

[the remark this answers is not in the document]
kmrdhruv: and these would be handled in AuthzHandler independent of authz provider/adapter ?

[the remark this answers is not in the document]
aayustark007-fk: yes, so anyone in superUsers list can bypass all checks

kmrdhruv (approved):
LGTM

PR #46, opened 2023-09-06
add producer level metrics

kmrdhruv:
TODO::
UT & E2E for produce path will be added post initial review.


common/src/main/java/com/flipkart/varadhi/AsyncResult.java
```diff
@@ -0,0 +1,37 @@
+package com.flipkart.varadhi;
+
+public class AsyncResult<T> {
+    private T result;
+    private Throwable cause;
+
+    public static <T> AsyncResult<T> of(T result, Throwable cause) {
```
aayustark007-fk, line 7: Vert.x [Future](https://vertx.io/docs/apidocs/io/vertx/core/Future.html) has the same capabilities. Are we doing this because we don't want to couple the common module with vertex dependencies?

kmrdhruv: yes, limit Vertx integration and dependencies to only Rest layer.

messaging/src/main/java/com/flipkart/varadhi/produce/otel/ProducerMetricProvider.java
```diff
@@ -0,0 +1,56 @@
...
+        tags.add(Tag.of(TAG_NAME_PROJECT, context.getTopicContext().getProjectName()));
+        tags.add(Tag.of(TAG_NAME_TOPIC, context.getTopicContext().getTopicName()));
+        tags.add(Tag.of(TAG_NAME_IDENTITY, context.getRequestContext().getProduceIdentity()));
+        tags.add(Tag.of(TAG_NAME_HOST, context.getRequestContext().getServiceHost()));
+        return tags;
+    }
+
+    private Counter getProducedBytesCounter(List<Tag> tags) {
+        // meter name - "produce.bytes", tags- region, project, topic, host
+        return Counter.builder(BYTES_METER).tags(tags).register(meterRegistry);
```
gauravAshok, line 47: do we have to register this counter every time whenever this method is called?
Having host in the metric tag/name has caused explosion in the total metrics in the past. We actually removed hosts from most places.

messaging/src/main/java/com/flipkart/varadhi/produce/otel/ProducerMetricProvider.java
```diff
@@ -0,0 +1,56 @@
...
+public class ProducerMetricProvider {
+    private final MeterRegistry meterRegistry;
+    private boolean enabled;
+
+    public ProducerMetricProvider(boolean enabled, MeterRegistry meterRegistry) {
+        this.meterRegistry = meterRegistry;
+        this.enabled = enabled;
+    }
+
+    public void OnMessageProduced(boolean succeeded, long producerLatency, ProduceContext context) {
```
gauravAshok, line 26:
```suggestion
    public void onMessageProduced(boolean succeeded, long producerLatency, ProduceContext context) {
```

messaging/src/main/java/com/flipkart/varadhi/produce/otel/ProducerMetricProvider.java
```diff
@@ -0,0 +1,56 @@
...
+        if (enabled) {
```
gauravAshok, line 27: usually, this enabled thing can be facilitated by interface & dummy impl.

messaging/src/main/java/com/flipkart/varadhi/produce/services/ProducerService.java
```diff
@@ -1,100 +1,81 @@
...
+            }
+            return AsyncResult.of(result, throwable);
+        });
+    }
+
+    private void emitProducerMetric(String messageId, boolean succeeded, int produceLatency, ProduceContext context) {
         try {
-            metricProvider.OnProduceEnd(messageId, status, produceLatency, context);
+            metricProvider.OnMessageProduced(succeeded, produceLatency, context);
         } catch (Exception e) {
```
gauravAshok, line 71: we shouldn't even catch "any" exception here.

server/src/main/java/com/flipkart/varadhi/config/VaradhiOptions.java
```diff
@@ -3,8 +3,15 @@
 import com.flipkart.varadhi.produce.config.ProducerOptions;
 import lombok.Data;
 
+import static com.flipkart.varadhi.Constants.REST_DEFAULTS.*;
+
 @Data
 public class VaradhiOptions {
```
gauravAshok, line 9: slightly weird name "VaradhiOptions" and also too broad. There is already a ServerConfiguration. that having "varadhiOption" does not make too much sense. For me Server is at same level as Varadhi, server just being the common name to refer to it.

server/src/main/java/com/flipkart/varadhi/web/v1/produce/HeaderValidationHandler.java
```diff
@@ -0,0 +1,53 @@
...
+
+public class HeaderValidationHandler {
+    private int headerNameSizeMax;
+    private int headerValueSizeMax;
+    private int headersAllowedMax;
+
+    public HeaderValidationHandler(VaradhiOptions varadhiOptions) {
+        this.headerNameSizeMax = varadhiOptions.getHeaderNameSizeMax();
+        this.headerValueSizeMax = varadhiOptions.getHeaderValueSizeMax();
+        this.headersAllowedMax = varadhiOptions.getHeadersAllowedMax();
```
gauravAshok, line 21: are these validations really required? Seems like an unnecessary overhead just to do this.

server/src/main/java/com/flipkart/varadhi/web/v1/produce/ProduceHandlers.java
```diff
@@ -89,8 +97,8 @@ public void produce(RoutingContext ctx) {
                                     ctx.endRequestWithResponse(produceResult.getMessageId());
                                 } else {
                                     ctx.endRequestWithStatusAndErrorMsg(
-                                            getHttpStatusForProduceStatus(produceResult.getProduceStatus().status()),
```
gauravAshok, line 92: what happens if we dont run this whenComplete handler using runOnContext()?

server/src/main/resources/log4j2.xml
```diff
@@ -6,7 +6,7 @@ TODO:: This is not being auto picked. Needs to be fixed.
   Preferring Log4j2 over logback primarily for Async logger feature.
   However detailed evaluation and configuration for Async loggers by default  yet to be done.
  -->
-<Configuration status="info">
+<Configuration produceStatus="info">
```
gauravAshok, line 9: bad rename?

server/src/main/resources/configuration.yml
```diff
@@ -43,3 +44,4 @@ httpServerOptions:
   alpnVersions: [ "HTTP_1_1", "HTTP_2" ]
   decompressionSupported: false
   useAlpn: true
+  tracingPolicy: "ALWAYS"
```
gauravAshok, line 47: where is this tracing being used?

kmrdhruv: This is for HTTP tracing in Vertx http server. This will force there is trace created for every http call (including produce).  If there is an existing trace, it will be used instead of creating new one.

server/src/main/java/com/flipkart/varadhi/web/v1/produce/HeaderValidationHandler.java
kmrdhruv: agree on performance part.. but it ensures headers are not used as mechanism to dump arbitrary info.

messaging/src/main/java/com/flipkart/varadhi/produce/otel/ProducerMetricProvider.java
kmrdhruv: https://micrometer.io/docs/concepts#_global_registry
This will return an existing meter if it was already registered. For now avoiding to cache these as well. Other option is to first find and then create if not found. That seems more expensive.
There is alternative method meterregistry.Counter(name, tags), it also does the same as above, might be more readable ?

kmrdhruv: agree on host part might cause issues w.r.to cardinality, but it seems to be have value add for analysis purpose. With K8 it might become even more verbose. Not sure about it, currently it is set though.

messaging/src/main/java/com/flipkart/varadhi/produce/services/ProducerService.java
kmrdhruv: handle it inside metric provider or fail produce ?

common/src/main/java/com/flipkart/varadhi/AsyncResult.java
```diff
@@ -0,0 +1,37 @@
+package com.flipkart.varadhi;
+
+public class AsyncResult<T> {
+    private T result;
```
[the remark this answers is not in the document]
kmrdhruv: FailableResult ?

server/src/main/java/com/flipkart/varadhi/web/v1/produce/ProduceHandlers.java
kmrdhruv: Not sure, but I believe it should be fine as long as context is maintained. However, i guess runOnContext() or ctx.end() both seems to be ending up on event loop thread.

messaging/src/main/java/com/flipkart/varadhi/produce/otel/ProducerMetricProvider.java
kmrdhruv: double check if opentelemetry is adding the host tag by default.

PR #51, opened 2023-10-17
Replace Snake Yaml with Jackson object mapper using yaml factory

aayustark007-fk:
SnakeYaml is not able to deserialise generic types properly.
`List<Enum>` is deserialised as a type erased `List` having String elements which leads to class cast exceptions.
I guess this is because of type erasure, it is not able to infer the correct type once the code is compiled.

However, I found that Jackson does not face this issue. How it is able to maintain the type information and avoid this problem, that is not known. But, for now we can safely replace usage of snake yaml with Jackson object mapper using YAMLFactory extension.

Following example demonstrates the issue:

Let's say we have a Config class:
```java
public class DefaultAuthorizationConfiguration {
    private Map<String, List<ResourceAction>> roles;
}
```

And the following YAML config:
```yaml
---
roles:
  org.admin:
    - ORG_CREATE
    - ORG_UPDATE
    - ORG_GET
    - ORG_DELETE
    - TEAM_CREATE
    - TEAM_GET
    - TEAM_UPDATE
    - PROJECT_GET
    - TOPIC_GET
```

Using the Snake Yaml reader we get the following deserialisation result:
Screenshot 2023-10-17 at 2 59 15 PM
Note that the values of map keys are raw `ArrayList` having `String` elements. But, we were expecting the type to be `ResourceAction`

Now using Jackson YAMLFactory based reader we get:
Screenshot 2023-10-17 at 3 03 21 PM
Here we see that the elements of the list are deserialised with correct types.

Hence, the PR is raised to replace Snake Yaml with Jackson due to more predictable deserialisation results.

### DIFF
--- a/common/src/main/java/com/flipkart/varadhi/utils/YamlLoader.java
+++ b/common/src/main/java/com/flipkart/varadhi/utils/YamlLoader.java
@@ -1,16 +1,22 @@
 package com.flipkart.varadhi.utils;
 
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.dataformat.yaml.YAMLFactory;
 import com.flipkart.varadhi.exceptions.VaradhiException;
-import org.yaml.snakeyaml.Yaml;
 
-import java.nio.file.Files;
-import java.nio.file.Path;
+import java.io.File;
 
 public class YamlLoader {
 
+    private static final ObjectMapper mapper = new ObjectMapper(new YAMLFactory());
+
+    static {
+        mapper.findAndRegisterModules();
+    }
+
     public static <T> T loadConfig(String configFile, Class<T> clazz) {
         try {
-            return new Yaml().loadAs(Files.readString(Path.of(configFile)), clazz);
+            return mapper.readValue(new File(configFile), clazz);
         } catch (Exception e) {
             throw new VaradhiException(String.format("Failed to load config file: %s as %s.", configFile, clazz), e);
         }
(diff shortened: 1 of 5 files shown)


common/src/main/java/com/flipkart/varadhi/utils/YamlLoader.java
```diff
@@ -1,16 +1,22 @@
...
+import java.io.File;
 
 public class YamlLoader {
 
+    private static final ObjectMapper mapper = new ObjectMapper(new YAMLFactory());
+
+    static {
+        mapper.findAndRegisterModules();
+    }
+
```
kmrdhruv, line 16: Can you add a testcase which covers the generic use cases at nested level and top level.

kmrdhruv (changes requested):

common/src/main/java/com/flipkart/varadhi/utils/YamlLoader.java
aayustark007-fk: add the test cases
using snake yaml following failure happens:
```
java.lang.ClassCastException: class java.lang.String cannot be cast to class com.flipkart.varadhi.utils.YamlLoaderTest$MyEnum (java.lang.String is in module java.base of loader 'bootstrap'; com.flipkart.varadhi.utils.YamlLoaderTest$MyEnum is in unnamed module of loader 'app')
```

but with jackson test passes

kmrdhruv (approved):
LGTM

PR #52, opened 2023-10-27
Changes to allow external AuthZ providers to integrate

aayustark007-fk:
Separate out the Authorization interfaces to separate module.
Add maven publish plugin to allow uploading artefacts to maven.
Before publishing add the remote repo url to `gradle.properties` for `mavenPublishUrl`


authz/src/test/java/com/flipkart/varadhi/authz/DefaultAuthorizationProviderTest.java
```diff
@@ -197,28 +198,42 @@ public void testIsAuthorized_UserProjectAccess(VertxTestContext testContext) {
...
+                .compose(t -> defaultAuthorizationProvider.isAuthorized(testUser("proj_user1", false),
                         ResourceAction.PROJECT_GET, "flipkart/team_rocket/proj001"
                 )) // proj_user1 is project.read so should work
                 .onComplete(testContext.succeeding(t -> {
                     Assertions.assertTrue(t);
                     checkpoint.flag();
                 }));
     }
+
+    public static UserContext testUser(String name, boolean expired) {
```
kmrdhruv, line 226: Might be better to put into testFixtures, can possibly used by other modules as well ?

buildSrc/src/main/groovy/com.flipkart.varadhi.java-library-conventions.gradle
```diff
@@ -4,4 +4,29 @@ plugins {
...
+}
+
+
+publishing {
+    publications {
+        maven(MavenPublication) {
+            groupId = "${projectGroupId}"
+            artifactId = "${project.name}"
+            version = "${version}"
+            from components.java
```
kmrdhruv, line 19: What all artifects/components will get published with this ?

buildSrc/src/main/groovy/com.flipkart.varadhi.java-library-conventions.gradle
```diff
@@ -4,4 +4,29 @@ plugins {
...
+    id 'maven-publish'
+}
+
+
+publishing {
+    publications {
+        maven(MavenPublication) {
+            groupId = "${projectGroupId}"
+            artifactId = "${project.name}"
+            version = "${version}"
```
kmrdhruv, line 18: How will this result in snapshot or unstable version being published ?

buildSrc/src/main/groovy/com.flipkart.varadhi.java-library-conventions.gradle
```diff
@@ -4,4 +4,29 @@ plugins {
 
     // Apply the java-library plugin for API and implementation separation.
     id 'java-library'
+
+    // Apply the maven-publish to allow deploying to maven artifactories
+    id 'maven-publish'
+}
+
+
+publishing {
```
kmrdhruv, line 13: This will add publishing task to each subproject ? Not all libraries needs to be published ?

buildSrc/src/main/groovy/com.flipkart.varadhi.java-library-conventions.gradle
```diff
@@ -4,4 +4,29 @@ plugins {
...
+
+            pom {
+                name = "${project.name}"
+            }
+        }
+    }
+
+    repositories {
+        maven {
+            url = mavenPublishUrl.replace('"', '') // important to strip double-quotes from this otherwise the url will be dealt as a file url
```
kmrdhruv, line 29: yes, let's get this fixed as well.

buildSrc/src/main/groovy/com.flipkart.varadhi.java-library-conventions.gradle
```diff
@@ -4,4 +4,29 @@ plugins {
...
+publishing {
+    publications {
+        maven(MavenPublication) {
+            groupId = "${projectGroupId}"
+            artifactId = "${project.name}"
+            version = "${version}"
+            from components.java
+
+            pom {
+                name = "${project.name}"
```
kmrdhruv, line 22: Do we need other details as well here ?

authz/build.gradle
```diff
@@ -0,0 +1,17 @@
+plugins {
```
kmrdhruv, line 1: would it make sense to move entities/auth as well into this module ?

kmrdhruv (changes requested):

authz/build.gradle
aayustark007-fk: if modules are owning their entities then it would make the entities module redundant

authz/src/test/java/com/flipkart/varadhi/authz/DefaultAuthorizationProviderTest.java
aayustark007-fk: sure will do that

buildSrc/src/main/groovy/com.flipkart.varadhi.java-library-conventions.gradle
aayustark007-fk: any module using java-library-conventions will get published
shall we can create a separate convention extending this for publishable-libraray-convention?

buildSrc/src/main/groovy/com.flipkart.varadhi.java-library-conventions.gradle
aayustark007-fk: won't be published unless gradlew publish is run
I have published SNAPSHOT version, but pushed changes have omitted SNAPSHOT since the docker build was failing

buildSrc/src/main/groovy/com.flipkart.varadhi.java-library-conventions.gradle
aayustark007-fk: authz, commons and entities

buildSrc/src/main/groovy/com.flipkart.varadhi.java-library-conventions.gradle
aayustark007-fk: this is to generate pom.xml so that the jar plays well when including in maven projects

aayustark007-fk: we can add more details if required, but currently I don't think anything else is required

buildSrc/src/main/groovy/com.flipkart.varadhi.java-library-conventions.gradle
aayustark007-fk: done

authz/src/test/java/com/flipkart/varadhi/authz/DefaultAuthorizationProviderTest.java
aayustark007-fk: done

authz/build.gradle
kmrdhruv: agree for Varadhi domain entities, but entities/auth has enum like resource types and resource actions and config which seems different from other entities. It has no intended uses in other modules (except authz and server).

buildSrc/src/main/groovy/com.flipkart.varadhi.java-library-conventions.gradle
kmrdhruv: Yes, either separate publishing convention or add it specifically to the required modules. May be former is better.

buildSrc/src/main/groovy/com.flipkart.varadhi.java-library-conventions.gradle
kmrdhruv: I think in code it should be SNAPSHOT by default. Why docker build was failing ?
(what I meant, is adding publishing to common convention will make that target visible to all submodules. Someone running publish at root, will execute it for all modules which we can avoid).

buildSrc/src/main/groovy/com.flipkart.varadhi.java-library-conventions.gradle
kmrdhruv: What I was asking is w.r.to  "components.java" ? What all does it publish for a single library module.

kmrdhruv (changes requested):
added few comments for clarification.

buildSrc/src/main/groovy/com.flipkart.varadhi.java-library-conventions.gradle
aayustark007-fk: docker build is hardcoded to use specific verison
I'll fix that

buildSrc/src/main/groovy/com.flipkart.varadhi.java-library-conventions.gradle
aayustark007-fk: not sure about this, I saw it was provided as part of publish syntax
published jar has to be maven compatible so it adds the required metadata to be able to correctly indexed by maven repos

buildSrc/src/main/groovy/com.flipkart.varadhi.java-library-conventions.gradle
aayustark007-fk: created separate publish config

authz/build.gradle
aayustark007-fk: done

setup/docker/Dockerfile
```diff
@@ -14,7 +14,7 @@ EXPOSE 8080
 RUN mkdir -p /usr/share/varadhi/dependencies && chmod 755 /usr/share/varadhi/dependencies
 RUN mkdir /etc/varadhi && chmod 755 /etc/varadhi
 COPY $VARADHI_HOME/server/build/dependencies/ /usr/share/varadhi/dependencies/
-COPY $VARADHI_HOME/pulsar/build/libs/pulsar-0.1.0.1.jar /usr/share/varadhi/dependencies/
+COPY $VARADHI_HOME/pulsar/build/libs/pulsar-*.jar /usr/share/varadhi/dependencies/
```
kmrdhruv, line 17: Why is this needed ?

kmrdhruv (approved):
Looks fine. Let's verify what all gets published. Not sure, iirc, it had lot more artifacts as part of publish, see if it is required.

PR #64, opened 2023-12-13
Made entities the base module.

### LINKED ISSUE #62: Initial Refactor to make entities a base module.
Currently Entities module bring in lot of dependency such as micrometer, common module, jackson, grizzly.

Since the name is entities, it should be the top-most and dependency free module of all. This module also can published.

If we are looking for entity ser-de, then it can be provided as part of another module, which only brings jackson dependency and allows entities ser-de.

There is common module as well, which contains some common functionality, for the varadhi project. The common module probably does not need publishing. Hence it can take depedency on entities (if required).

gauravAshok:
- Cleaned up dependencies. 
- Entities will be api dependency for other modules.
- Common module now depends on entities. Earlier it was opposite.
- Entities now only explicitly depends on jackson-annotations, jakarta-api as api dependency.
- Moved some of the constants to the core related classes. initial_version -> base resource. name_delimiter to AbstractTopic.
- Removed exception classes which are duplicate of standard java exceptions.

Resolves #62

### DIFF
--- a/entities/build.gradle
+++ b/entities/build.gradle
@@ -5,7 +5,6 @@ dependencies {
     api("com.fasterxml.jackson.core:jackson-annotations:$jackson_version")
     api("jakarta.validation:jakarta.validation-api:$jakarta_version")
 
-    testRuntimeOnly("jakarta.el:jakarta.el-api:4.0.0")
-    testRuntimeOnly("org.glassfish:jakarta.el:4.0.2")
+    testRuntimeOnly("org.glassfish.expressly:expressly:5.0.0")
     testRuntimeOnly("org.hibernate:hibernate-validator:$hibernate_validator_version")
 }
--- a/entities/build.gradle
+++ b/entities/build.gradle
@@ -1,12 +1,11 @@
 plugins {
-    id "com.flipkart.varadhi.java-library-conventions"
+    id "com.flipkart.varadhi.java-published-library-conventions"
 }
 dependencies {
-    implementation(project(":common"))
-    implementation("io.micrometer:micrometer-core")
-    implementation("com.fasterxml.jackson.core:jackson-databind")
-    implementation("jakarta.validation:jakarta.validation-api")
-    implementation('org.hibernate:hibernate-validator:8.0.1.Final')
-    implementation('org.glassfish.expressly:expressly:5.0.0')
-    testImplementation(project(":pulsar"))
+    api("com.fasterxml.jackson.core:jackson-annotations:$jackson_version")
+    api("jakarta.validation:jakarta.validation-api:$jakarta_version")
+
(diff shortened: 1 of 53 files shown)


entities/src/main/java/com/flipkart/varadhi/entities/AbstractTopic.java
```diff
@@ -1,6 +1,9 @@
 package com.flipkart.varadhi.entities;
 
 public class AbstractTopic extends VaradhiResource {
+
+    public static String NAME_SEPARATOR = ".";
```
kmrdhruv, line 5: Better to move to VaradhiResource, other entities likely to have similar requirement e.g. subscription/queue.

entities/src/main/java/com/flipkart/varadhi/entities/VaradhiTopic.java
```diff
@@ -43,11 +41,14 @@ public void addInternalTopic(InternalTopic internalTopic) {
...
     }
 
-    public InternalTopic getProduceTopicForRegion(String region) {
-        InternalTopic internalTopic = internalTopics.get(region);
-        if (null == internalTopic) {
-            throw new ResourceNotFoundException(String.format("Topic not found for region(%s).", region));
-        }
-        return internalTopic;
+    /**
+     * TODO: evaluate whether it should return Optional or not. An unmapped region probably wouldn't be queried.
```
kmrdhruv, line 45: agree, unmapped region shouldn't be queried. If it does' letting it fail is better.
Optional seems additional overhead, as caller need to do the additional validations on top of it, so can be removed.

entities/build.gradle
```diff
@@ -1,12 +1,10 @@
...
 dependencies {
-    implementation(project(":common"))
-    implementation("io.micrometer:micrometer-core")
-    implementation("com.fasterxml.jackson.core:jackson-databind")
-    implementation("jakarta.validation:jakarta.validation-api")
-    implementation('org.hibernate:hibernate-validator:8.0.1.Final')
-    implementation('org.glassfish.expressly:expressly:5.0.0')
-    testImplementation(project(":pulsar"))
+    api("com.fasterxml.jackson.core:jackson-annotations:$jackson_version")
+    api("jakarta.validation:jakarta.validation-api:$jakarta_version")
```
kmrdhruv, line 6: why api ?

kmrdhruv (changes requested):

entities/build.gradle
gauravAshok: synced offline.
simple rule to follow: if the class is referable due to the exposed methods, then it is part of api.

kmrdhruv (approved):
LGTM

PR #68, opened 2023-12-19
Adding subscription related entities.

### LINKED ISSUE #63: Add subscription and other consumer specific entities.

gauravAshok:
resolves #63.

So I have noticed few things and I would like to place some guidelines in how we are going to define entities.

* The entities module is for entities that will be included in our "apis". Since varadhi will have a lot of pluggables, via spi, the service provider entities are also part of the api. The entities in the http layer is also part of the api.
* We may define multiple classes which may be related to one core entity, for eg: a subscription. But these classes needs to be part of some api for it to be in entity module. We will still try to reduce such multiplicity / duplications.
* 2 main candidates for apis are storage layer & web layer. We can expect 2 classes (at most) for any entity for the mentioned apis.
* Entities for storage layer should be lean and free of redundancy & complexity. Dont have nullable fields if possible.
* Entities for rest-layer needs to be rest api centric. i.e. We can think of not having uber updateSubscription api, and instead have targeted apis. updateConsumptionPolicy() api. updateEndpoint() api.. and so on. We may choose to use same pojo for at rest layer & storage layer, if the pojo is fundamental in nature. But having a rest specific entities may be more flexible, as we can have nullable fields there.
* Entities only required internally in other modules can be in their respective modules.

### DIFF
--- a/entities/src/main/java/com/flipkart/varadhi/entities/Endpoint.java
+++ b/entities/src/main/java/com/flipkart/varadhi/entities/Endpoint.java
@@ -1,0 +1,32 @@
+package com.flipkart.varadhi.entities;
+
+import jakarta.ws.rs.HttpMethod;
+import lombok.Data;
+import lombok.EqualsAndHashCode;
+
+import java.net.URL;
+
+public abstract sealed class Endpoint {
+
+    abstract Protocol getProtocol();
+
+    @EqualsAndHashCode(callSuper = true)
+    @Data
+    public static final class HttpEndpoint extends Endpoint {
+        private final URL url;
+        private final HttpMethod method;
+        private final String contentType;
+
+        private final boolean http2Supported;
+
+        @Override
+        Protocol getProtocol() {
+            return http2Supported ? Protocol.HTTP2 : Protocol.HTTP1_1;
+        }
+    }
+
+    public enum Protocol {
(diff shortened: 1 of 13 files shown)


entities/src/main/java/com/flipkart/varadhi/entities/Endpoint.java
```diff
@@ -0,0 +1,32 @@
...
+public abstract sealed class Endpoint {
+
+    abstract Protocol getProtocol();
+
+    @EqualsAndHashCode(callSuper = true)
+    @Data
+    public static final class HttpEndpoint extends Endpoint {
+        private final URL url;
+        private final HttpMethod method;
+        private final String contentType;
```
kmrdhruv, line 18: Is this for default value in case it is not available with message ? If yes, should it be with topic instead of subscription ?

kmrdhruv (approved):
LGTM.
check for comment on contentType in endpoint.

entities/src/main/java/com/flipkart/varadhi/entities/Endpoint.java
gauravAshok: Hmm.. I actually didn't put much thought here :)
Topic defining the content-type makes complete sense.

Will remove in the future PR for now.

PR #77, opened 2024-01-02
Base Entities changes

aayustark007-fk:
For core entities, like Topic and Subscription, we are following the convention:
- [entity]Resource: User facing CRUD contract, not persisted, versioned
- Varadhi[entity]: Internal entity, persisted, versioned

For management entities like, Org, Team, Project, same class can be used for both.

However, to achieve the above we need separation of persisted entity and versioned entity. Currently, VaradhiResource is the one acting as both.

This PR splits that into: MetaStoreEntity (persistence) and VersionedEntity (versioning). Also, renamed BaseEntity to Validatable. Resources only extend VersionedEntity while the Varadhi[entity] extends the MetaStoreEntity.

Class hierarchy now looks like:

Screenshot 2024-01-05 at 4 49 38 PM

### DIFF
--- a/entities/src/main/java/com/flipkart/varadhi/entities/Subscription.java
+++ b/entities/src/main/java/com/flipkart/varadhi/entities/Subscription.java
@@ -2,7 +2,7 @@ package com.flipkart.varadhi.entities;
 
 import jakarta.validation.constraints.NotNull;
 
-public class Subscription extends VaradhiResource {
+public class Subscription extends MetaStoreEntity {
     @NotNull
     private String topic;
     private boolean grouped;
--- a/entities/src/main/java/com/flipkart/varadhi/entities/TopicResource.java
+++ b/entities/src/main/java/com/flipkart/varadhi/entities/TopicResource.java
@@ -5,8 +5,8 @@ import lombok.Value;
 
 @Value
 @EqualsAndHashCode(callSuper = true)
-@ValidateVaradhiResource(message = "Invalid Topic name. Check naming constraints.", max = 64)
-public class TopicResource extends VaradhiResource {
+@ValidateMetaStoreEntity(message = "Invalid Topic name. Check naming constraints.", max = 64)
+public class TopicResource extends MetaStoreEntity {
     private static final String RESOURCE_TYPE_NAME = "TopicResource";
 
     String project;
--- a/entities/src/main/java/com/flipkart/varadhi/entities/TopicResource.java
+++ b/entities/src/main/java/com/flipkart/varadhi/entities/TopicResource.java
@@ -5,8 +5,9 @@ import lombok.Value;
 
 @Value
 @EqualsAndHashCode(callSuper = true)
-@ValidateMetaStoreEntity(message = "Invalid Topic name. Check naming constraints.", max = 64)
-public class TopicResource extends MetaStoreEntity {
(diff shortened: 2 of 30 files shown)


entities/src/main/java/com/flipkart/varadhi/entities/Subscription.java
```diff
@@ -2,7 +2,7 @@
 
 import jakarta.validation.constraints.NotNull;
 
-public class Subscription extends VaradhiResource {
+public class Subscription extends MetaStoreEntity {
```
kmudgal25, line 5: Why are we not using the validation annotation here?

entities/src/main/java/com/flipkart/varadhi/entities/TopicResource.java
```diff
@@ -5,8 +5,8 @@
 
 @Value
 @EqualsAndHashCode(callSuper = true)
-@ValidateVaradhiResource(message = "Invalid Topic name. Check naming constraints.", max = 64)
-public class TopicResource extends VaradhiResource {
+@ValidateMetaStoreEntity(message = "Invalid Topic name. Check naming constraints.", max = 64)
+public class TopicResource extends MetaStoreEntity {
```
kmudgal25, line 9: shouldn't this extend VersionedEntity as it is user facing entity?

kmudgal25 (changes requested):

entities/src/main/java/com/flipkart/varadhi/entities/Subscription.java
aayustark007-fk: this will get removed in the subscription crud PR

entities/src/main/java/com/flipkart/varadhi/entities/TopicResource.java
aayustark007-fk: Topic refactoring to conform to this model will happen as separate PR. I think that PR is being already worked on.

entities/src/main/java/com/flipkart/varadhi/entities/auth/RoleBindingNode.java
```diff
@@ -15,8 +15,8 @@
 
 @Getter
 @EqualsAndHashCode(callSuper = true)
-@ValidateVaradhiResource(message = "Invalid Role Binding name. Check naming constraints.")
-public class RoleBindingNode extends VaradhiResource {
+@ValidateResource(message = "Invalid Role Binding name. Check naming constraints.")
+public class RoleBindingNode extends MetaStoreEntity {
```
kmrdhruv, line 19: Is it validatable or not as it has all constraints in-place ?
If it is, then mostly default values of length wouldn't suffice as resourceId is concatenation of multiple components.

kmrdhruv (changes requested):
some clarification on RoleBindingNode.

entities/src/main/java/com/flipkart/varadhi/entities/auth/RoleBindingNode.java
aayustark007-fk: rolebinding node only extends metastore as it is not user facing on the request side
rather I have made IAMPolicyUpdateRequest implement Validatable

kmrdhruv (approved):
LGTM

PR #79, opened 2024-01-03
Vertx

### LINKED ISSUE #101: Enable metric collection via jmx / otel

mohitkumarv76:
Jmx Metrics when enabled for a timed api, currently looks like this:

`CreateTopicLatency.resource./public/team_rocket/default/test-topic-1` 

closes #101


gauravAshok (commented):
I think I am more or less aligned with the approach. We are able to customize the VertxMetrics to use our own HttpServerMetrics impl. There we can use HttpRequest & HttpResponse object to emit whatever metrics we want to emit.

Few doubts:
- What is the role of ResourceName? What is its purpose?
- I am guessing the current content of CustomHttpServerMetrics is only a POC and not a final approach. So, what would be final approach? I dont expect any metrics to emitted during `requestBegin`, but only during `responseEnd`. Also, currently due to shear number of tags, I am expecting a lot of metrics to be created. So can we rationalize the actual tags required? Can we judge the number of unique metrics that will be created due to tag combinations?
-

mohitkumarv76:
Not able to reply in above comment, updating questions here :
What is the role of ResourceName? What is its purpose? ->. I have updated it to metric name for clarity, through it we can ask user against which metric name dev wants to push for the metrics emitted by the api.
For second -> It is final approach, if we don't want to handle requestBegin that is our option. HttpServerMetrics interface provides us many method at which we will get control & can push metrics. Regarding tags had meet with Dhruv this we will handle later, scope of this issue was only to publish metrics. We can connect later and finalise what we want to capture and along with which tags & add those. @gauravAshok

server/build.gradle
```diff
@@ -30,6 +30,7 @@ dependencies {
     implementation("io.vertx:vertx-opentelemetry")
     implementation("io.vertx:vertx-opentracing")
     implementation("io.vertx:vertx-micrometer-metrics")
+    implementation("io.micrometer:micrometer-registry-prometheus")
```
gauravAshok, line 33: Why add this?

server/src/main/java/com/flipkart/varadhi/metrices/MetricConstants.java
```diff
@@ -0,0 +1,12 @@
+package com.flipkart.varadhi.metrices;
+
+public class MetricConstants {
+
+    public static final String INPUT_REQUEST = "http_input_request";
+    public static final String TOTAL_REQUEST_COUNT = "http_request_count";
+    public static final String TOTAL_FAILURE_COUNT = "http_failure_count";
+    public static final String TOTAL_SUCCESS_COUNT = "http_success_count";
+    public static final String OUTPUT_RESPONSE_CODE = "http_output_response";
```
gauravAshok, line 9: TOTAL_REQUEST_COUNT & OUTPUT_RESPONSE_CODE looks duplicate in nature.

PR #100, opened 2024-02-08
Adding consumer spi, varadhi consumer, consumer manager interfaces

### LINKED ISSUE #72: Add a consumer-core module. Add Subscriber entity. Subscriber listens to Subscription updates & lifecycle events (start, stop, close).
This is the core of the consumer. It will be responsible for message processing and delivery.
Other modules will be responsible for packaging this as a library or as a server.

### LINKED ISSUE #73: Add node registeration, subscription-control events. Manages multiple subscriber instances based on the control events.
This can be added in the server module itself.

gauravAshok:
Closes #72 
Closes #73 

#73 talks about node registeration, but that is already taken care of in another open PR.

- What is the identifier of a subscription? It is kind of needed to refer to it when we want to stop it or ask info about it. Check ConsumersManager.
- I have kept ConsumersManager in server module only for now. If it remains largely independent of server workings, then we can put it in consumer module in future refactoring.
- The SubscriptionShard info is looking like a separate entity from core Subscription entity. During startSubscription operation, the params are core sub entity and one of its shard, but right now since shard is inside the Subscription entity, the startSubscription operation is only taking Subscription entity and putting the constraint that it must only contain one shard info, which is wierd.


spi/src/main/java/com/flipkart/varadhi/spi/services/Consumer.java
```diff
@@ -0,0 +1,36 @@
...
+     * @return
+     */
+    CompletableFuture<Void> commitCummulative(String topic, int partition, O offset);
+
+    /**
+     * Unsubscribes the topics and closes the consumer object and any associated resources.
+     */
+    void close();
+
+    // TODO: evaluate pause and resume methods.
```
gauravAshok, line 35: most likely not going to be there.

spi/src/main/java/com/flipkart/varadhi/spi/services/PartitionMessages.java
```diff
@@ -0,0 +1,17 @@
+package com.flipkart.varadhi.spi.services;
+
+import com.flipkart.varadhi.entities.Offset;
+import lombok.AllArgsConstructor;
+
+/**
+ * Represent a batch of messages polled from a particular partition.
+ */
+@AllArgsConstructor
+public class PartitionMessages<O extends Offset> {
```
kmrdhruv, line 10: 1. will below placed in spi or entities ?
2. should paritition be abstracted as well or how it will be converted into tech stack specific identifier e.g. in pulsar partitioned topic has its own identifier  ?

spi/src/main/java/com/flipkart/varadhi/spi/services/PolledMessage.java
```diff
@@ -0,0 +1,23 @@
...
+
+    /**
+     * @return the offset of this message in the partition.
+     */
+    O offset();
+
+    /**
+     * @return the payload of this message.
+     */
+    byte[] payload();
```
kmrdhruv, line 20: what about headers ? should it be message object instead with those semantics attached ?

kmrdhruv (changes requested):

consumer/src/main/java/com/flipkart/varadhi/consumer/ConsumerStatus.java
```diff
@@ -0,0 +1,18 @@
...
+    CONSUMING,
+
+    ERRORED,
+
+    /**
+     * Throttled is a wide bucket. TODO: evaluate splitting based on the requirements later on.
+     */
+    THROTTLED,
+
+    PAUSED
```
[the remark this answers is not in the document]
gauravAshok: I actually think "pause" meaning stop message delivery immediately as a nice feature. It has nothing to do with allocation. 
Basically an api, that you can use to "pause" message delivery and which works immediately.

we can also think of downgrading extended "pause" duration to "stopped" state.

consumer/src/main/java/com/flipkart/varadhi/consumer/VaradhiConsumer.java
```diff
@@ -0,0 +1,38 @@
...
+
+import com.flipkart.varadhi.entities.VaradhiSubscription;
+
+/**
+ * Base varadhi consumer to consume messages from a varadhi topic. It corresponds to a single shard of some
+ * subscription.
+ */
+public interface VaradhiConsumer {
+
+    VaradhiSubscription getSubscription();
```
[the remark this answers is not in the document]
gauravAshok: ya. this will be reworked to point to a shard. but getter will be there. it is a consumer for a subscription, so a pure getter to again fetch the subscirption details for whcih this consumer was created seems useful. Also, the same can probably be used to check whether the consumer is created for a subscription with an old version.

If it turns out that the this method is not useful, we can remove it.

server/src/main/java/com/flipkart/varadhi/services/ConsumersManager.java
```diff
@@ -0,0 +1,31 @@
...
+
+    Future<Void> stopSubscription(String subscription, int shardId);
+
+    Future<Void> pauseSubscription(String subscription, int shardId);
+
+    Future<Void> resumeSubscription(String subscription, int shardId);
+
+    ConsumerStatus getConsumerStatus(String subscription, int shardId);
+
+    // TODO likely need status on the starting / stopping as well; as the above status is for a running consumer..
```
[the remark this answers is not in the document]
gauravAshok: didn't follow exactly.

server/src/main/java/com/flipkart/varadhi/services/ConsumersManager.java
```diff
@@ -0,0 +1,31 @@
...
+     * Start consumption for a shard of a subscription. If subscription failed, the future will contain the exception
+     * details.
+     *
+     * @param subscription
+     *
+     * @return
+     */
+    Future<Void> startSubscription(VaradhiSubscription subscription);
+
+    Future<Void> stopSubscription(String subscription, int shardId);
```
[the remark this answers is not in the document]
gauravAshok: No it does not need to be indentified globally. 
This needs some id unique under the context of subscription. If the subscription object is global then sure the shard id together with sub id becomes globally unique automatically.

But as mentioned in the description we probably need subscription identifier to be able to use in inter-node communication. But this identifier does not have to be the identifier that is being used in the api layer.

spi/src/main/java/com/flipkart/varadhi/spi/services/PolledMessage.java
gauravAshok: I dont remember, but did we have some discussion around this before?

spi/src/main/java/com/flipkart/varadhi/spi/services/PartitionMessages.java
gauravAshok: `will below placed in spi or entities ?` - will evaluate.

gauravAshok: Can we keep this in spi itself? I am thinking that this class exists due to spi interfaces. If the interfaces are not there, then there is no need for this entity itself. This entity pertains to spi interfaces.

And about the partition abstraction, I think this particular works for kafka & pulsar at least. In case of non-partitioned scenario, the partition id can be -1. and the topic can refer to the relevant reference.

I am also thinking of platforms that dont even have partitioned concept. To support them in future, I think instead of coming up with 1 general interface that somehow fits all usecases, we can broadly define 2 feature capabalities. Partitioned vs non-partitioned.
partitioned stack like (kafka, pulsar google pub-sub lite) can fit in the model present in this PR.
whereas a non-partitioned stack (whcih also will require a different sharding model) can use a different set of interface.

This feature capability value will obviously come from the messagingStackProvider.

spi/src/main/java/com/flipkart/varadhi/spi/services/PolledMessage.java
gauravAshok: should I just refer entities.Message here instead of this payload array?

consumer/src/main/java/com/flipkart/varadhi/consumer/ConsumerStatus.java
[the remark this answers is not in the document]
gauravAshok: `subscription is running, but not consuming as there is nothing to consume` - it will be CONSUMING state.
`CONSUMING` is a "state" in which the consumer is in. When in "Consuming", the varadhi consumer will be attempting message delivery for un-delivered messages.
But if there are no undelivered messages, then obviously the destination wont get any. The consume rate can be fetched orthogonally from the stats exposed by the varadhi consumer.  

PAUSED - means that consumer is intentionally not delivering any messages.

spi/src/main/java/com/flipkart/varadhi/spi/services/PartitionMessages.java
gauravAshok: Or we dont rely on partitions at all. that will work in all cases.

kmrdhruv (approved):

PR #117, opened 2024-03-21
Cluster manager implementation and refactoring

### LINKED ISSUE #84: Implement cluster manager

kmrdhruv:
Closes #84 
1. Implements the VaradhiClusterManager by extending vertx zk cluster manager.
2. Membership change listener
3. refactoring in message channel implementation.


server/src/main/resources/configuration.yml
```diff
@@ -1,5 +1,9 @@
 
-components : ["Server", "Controller", "All"]
+member:
+  memberId: "app-1"
+  roles : ["Server", "Controller", "All"]
+  cpuCount: 1
```
gauravAshok, line 5: cpu count can be computed by the appliction.

server/src/main/java/com/flipkart/varadhi/cluster/MessageRouter.java
```diff
@@ -0,0 +1,64 @@
...
+
+
+    public MessageRouter(EventBus vertxEventBus, DeliveryOptions deliveryOptions) {
+        this.vertxEventBus = vertxEventBus;
+        this.deliveryOptions = deliveryOptions;
+    }
+
+    public <E extends ClusterMessage> void sendHandler(String routeName, String apiName, SendHandler<E> handler) {
+        vertxEventBus.consumer(getApiPath(routeName, apiName, RouteMethod.SEND), message -> {
+            E cm = (E) JsonMapper.jsonDeserialize((String) message.body(), ClusterMessage.class);
```
gauravAshok, line 43: we need to fix the jsonDeserialize to not serialize to string but to byte[].

server/src/main/java/com/flipkart/varadhi/cluster/VaradhiClusterManager.java
```diff
@@ -9,15 +11,15 @@
...
-    List<NodeInfo> getAllMembers();
+    Future<List<MemberInfo>> getAllMembers();
 
     void addMembershipListener(MembershipListener listener);
 
-    MessageChannel connect(String nodeId);
+    MessageRouter getRouter(Vertx vertx);
 
-    // TODO: Any publish to all methods?
+    MessageExchange getExchange(Vertx vertx);
```
gauravAshok, line 22: any reason to call it exchange?

from first reading, the difference between exchange and router is not that clear.
I see that exchange & router are apposite of each other; similar to producer vs consume.

core/src/main/java/com/flipkart/varadhi/core/cluster/ComponentKind.java
```diff
@@ -1,4 +1,4 @@
-package com.flipkart.varadhi.components;
+package com.flipkart.varadhi.core.cluster;
```
gauravAshok, line 1: should ALL be part of KIND?
or ALL should implicitly be represented by including actual kinds.

gauravAshok (approved):
LGTM

server/src/main/java/com/flipkart/varadhi/cluster/MessageExchange.java
```diff
@@ -0,0 +1,57 @@
...
+
+    public MessageExchange(EventBus vertxEventBus, DeliveryOptions deliveryOptions) {
+        this.vertxEventBus = vertxEventBus;
+        this.deliveryOptions = deliveryOptions;
+    }
+
+    public CompletableFuture<Void> send(String routeName, String apiName, ClusterMessage msg) {
+        CompletableFuture<Void> future = new CompletableFuture<>();
+        vertxEventBus.request(
+                getPath(routeName, apiName, RouteMethod.SEND), JsonMapper.jsonSerialize(msg), deliveryOptions, ar -> {
```
gauravAshok, line 34: so, here the path is vertx event bus "address". right?
So I still have concerns on this as the #addresses will increase with time. And the total #address to be managed by vertx / zk, will be proportional to api paths x node count.

I also have some doubts on the "addressing".

each web-server node will register the handler against the same path. so when the controller begins a request ("if" it begins), then the request can land on any web-server node. right? Just validating here.

Another concern is the ordering of events. If I send 2 events on 2 different "path", then I suppose there is no guarantee of the ordering between them. event_1 on path_1 is sent first and then event_2 on path_2 is sent without waiting for the response of event_1, then we probably cant assume anything about the ordering of these events. Dont think that this is an issue per say, but just pointing out.

gauravAshok: `will be proportional to api paths x node count.` : huh I may be wrong here. as the address dont have ip in them. (yet)

server/src/main/resources/configuration.yml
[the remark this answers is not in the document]
kmrdhruv: correct, this and other details can be auto calculated based on need.
Think of this as place holder for now, as resource specification will change further as subscription/shard assignment functionality is developed.

server/src/main/java/com/flipkart/varadhi/cluster/VaradhiClusterManager.java
kmrdhruv: Open for better names here.. 
Exchange vs Router is Produce vs Consume as you mentioned.

core/src/main/java/com/flipkart/varadhi/core/cluster/ComponentKind.java
kmrdhruv: it should be later.. i.e. no ALL.

server/src/main/java/com/flipkart/varadhi/cluster/MessageExchange.java
kmrdhruv: you are correct, number of watches would be approx equal to node count * api, but that's still bounded. 
open to re-work the address based on how they span post one or two flow implemented e2e.

Event ordering. -- aligned, i think we shouldn't rely on events on two independent path being ordered.

PR #134, opened 2024-04-17
Initial implementation for Grouped and UnGrouped Message Source

aayustark007-fk:
Implementation added for Grouped and UnGrouped Message source

UnGrouped messages source: Polls message batches from the consumer and populates the passed array. Extra messages are buffered in-mem in case of overflow. Subsequent calls will try to drain the buffer first.

Grouped message source: Polls one message per free groupId and populates the array. Further messages in a group will not be polled until the previous one gets committed.

This is still the initial implementation. Subject to change based on the final subscription consumer usage patterns.

### DIFF
--- a/consumer/src/main/java/com/flipkart/varadhi/consumer/GroupedMessageSrc.java
+++ b/consumer/src/main/java/com/flipkart/varadhi/consumer/GroupedMessageSrc.java
@@ -1,0 +1,160 @@
+package com.flipkart.varadhi.consumer;
+
+import com.flipkart.varadhi.entities.Message;
+import com.flipkart.varadhi.entities.Offset;
+import com.flipkart.varadhi.spi.services.Consumer;
+import com.flipkart.varadhi.spi.services.PolledMessage;
+import com.flipkart.varadhi.spi.services.PolledMessages;
+import lombok.AllArgsConstructor;
+import lombok.RequiredArgsConstructor;
+import org.apache.commons.lang3.StringUtils;
+import org.apache.commons.lang3.mutable.MutableBoolean;
+
+import java.util.*;
+import java.util.concurrent.BlockingDeque;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.LinkedBlockingDeque;
+import java.util.concurrent.atomic.AtomicLong;
+
+@RequiredArgsConstructor
+public class GroupedMessageSrc<O extends Offset> implements MessageSrc {
+
+    private final ConcurrentHashMap<String, GroupTracker> allGroupedMessages = new ConcurrentHashMap<>();
+    private final BlockingDeque<String> freeGroups = new LinkedBlockingDeque<>();
+    private final AtomicLong totalInFlightMessages = new AtomicLong(0);
+    private final long maxInFlightMessages = 100; // todo: make configurable
+
+    private final Consumer<O> consumer;
+
+    @Override
+    public CompletableFuture<Integer> nextMessages(MessageTracker[] messages) {
+        if (hasMaxInFlightMessages()) {
+            return replenishAvailableGroups().thenApply(v -> nextMessagesInternal(messages));
+        }
+        return CompletableFuture.supplyAsync(() -> nextMessagesInternal(messages));
+    }
+
+    private int nextMessagesInternal(MessageTracker[] messages) {
+        GroupTracker groupTracker = getGroupTracker();
+        if (null == groupTracker) {
+            return 0;
+        }
+
+        MessageTracker messageTracker = groupTracker.messages.getFirst().nextMessage();
+        messages[0] = new GroupedMessageTracker(messageTracker);
+        return 1;
+    }
+
+    private GroupTracker getGroupTracker() {
+        String freeGroup = freeGroups.poll();
+        if (freeGroup == null) {
+            return null;
+        }
+
+        GroupTracker tracker = allGroupedMessages.get(freeGroup);
+        if (tracker == null || tracker.status == GroupStatus.IN_FLIGHT) {
+            throw new IllegalStateException(String.format("Polled group %s: %s", freeGroup, tracker));
+        }
+
+        tracker.status = GroupStatus.IN_FLIGHT;
+        return tracker;
+    }
+
+    private CompletableFuture<Void> replenishAvailableGroups() {
+        return consumer.receiveAsync().thenApply(polledMessages -> {
+            replenishAvailableGroups(polledMessages);
+            return null;
+        });
+    }
+
+    private void replenishAvailableGroups(PolledMessages<O> polledMessages) {
+        Map<String, List<MessageTracker>> groupedMessages = groupMessagesByGroupId(polledMessages);
+        for (Map.Entry<String, List<MessageTracker>> group : groupedMessages.entrySet()) {
+            MessageBatch newBatch = new MessageBatch(group.getValue());
+            MutableBoolean isNewGroup = new MutableBoolean(false);
+            allGroupedMessages.compute(group.getKey(), (groupId, tracker) -> {
+                if (tracker == null) {
+                    tracker = new GroupTracker();
+                    isNewGroup.setTrue();
+                }
+                tracker.messages.add(newBatch);
+                return tracker;
+            });
+            totalInFlightMessages.addAndGet(newBatch.count());
+            if (isNewGroup.isTrue()) {
+                freeGroups.add(group.getKey());
+            }
+        }
+    }
+
+    private Map<String, List<MessageTracker>> groupMessagesByGroupId(PolledMessages<O> polledMessages) {
+        Map<String, List<MessageTracker>> groups = new HashMap<>();
+        for (PolledMessage<O> polledMessage : polledMessages) {
+            MessageTracker messageTracker = new PolledMessageTracker<>(consumer, polledMessage);
+            String groupId = messageTracker.getGroupId();
+            if (StringUtils.isBlank(groupId)) {
+                throw new IllegalStateException("Group id not found for message " + messageTracker.getMessage());
+            }
+            groups.computeIfAbsent(groupId, list -> new ArrayList<>()).add(messageTracker);
+        }
+        return groups;
+    }
+
+    private boolean hasMaxInFlightMessages() {
+        return totalInFlightMessages.get() >= maxInFlightMessages;
+    }
+
+    enum GroupStatus {
+        FREE,
+        IN_FLIGHT
+    }
+
+    static class GroupTracker {
+        GroupStatus status = GroupStatus.FREE;
+
+        LinkedList<MessageBatch> messages = new LinkedList<>();
+    }
+
+    @AllArgsConstructor
+    private class GroupedMessageTracker implements MessageTracker {
+        private final MessageTracker messageTracker;
+
+        @Override
+        public Message getMessage() {
+            return messageTracker.getMessage();
+        }
+
+        @Override
+        public void onConsumed(MessageConsumptionStatus status) {
+            messageTracker.onConsumed(status);
+            String groupId = getGroupId();
+            free(groupId, status);
+        }
+
+        private void free(String groupId, MessageConsumptionStatus status) {
+            MutableBoolean isRemaining = new MutableBoolean(false);
+            allGroupedMessages.compute(groupId, (gId, tracker) -> {
+                if (tracker == null || tracker.status == GroupStatus.FREE) {
+                    throw new IllegalStateException(String.format("Tried to free group %s: %s", gId, tracker));
+                }
+                // todo: update group consumption status in tracker?
+                var messages = tracker.messages;
+                if (!messages.isEmpty() && messages.getFirst().remaining() == 0) {
+                    messages.removeFirst();
+                }
+                if (!messages.isEmpty()) {
+                    tracker.status = GroupStatus.FREE;
+                    isRemaining.setTrue();
+                    return tracker;
+                } else {
+                    return null;
+                }
+            });
+            totalInFlightMessages.decrementAndGet();
+            if (isRemaining.isTrue()) {
+                freeGroups.addFirst(groupId);
+            }
+        }
+    }
+}
--- a/consumer/src/main/java/com/flipkart/varadhi/consumer/GroupedMessageSrc.java
+++ b/consumer/src/main/java/com/flipkart/varadhi/consumer/GroupedMessageSrc.java
@@ -122,7 +122,10 @@ public class GroupedMessageSrc<O extends Offset> implements MessageSrc {
     }
 
     enum GroupStatus {
+        // no message for this group is in active processing by the consumer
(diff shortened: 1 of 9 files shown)


consumer/src/main/java/com/flipkart/varadhi/consumer/UnGroupedMessageSrc.java
```diff
@@ -0,0 +1,50 @@
...
+            }
+        }
+        return i;
+    }
+
+    private int fetchFromBuffer(MessageTracker[] messages) {
+        int i = 0;
+        while (i < messages.length && !buffer.isEmpty()) {
+            messages[i++] = buffer.poll();
+        }
```
gauravAshok, line 47: Is there a better way to do it?
Instead of using linkedqueue, is there a way to use a arrayQueue and do a copy using a single call?

consumer/src/main/java/com/flipkart/varadhi/consumer/UnGroupedMessageSrc.java
```diff
@@ -0,0 +1,50 @@
...
+    private final BlockingQueue<MessageTracker> buffer = new LinkedBlockingQueue<>();
+
+    @Override
+    public CompletableFuture<Integer> nextMessages(MessageTracker[] messages) {
+        int offset = fetchFromBuffer(messages);
+        if (offset < messages.length) {
+            return consumer.receiveAsync()
+                    .thenApply(polledMessages -> processPolledMessages(offset, polledMessages, messages));
+        }
+        return CompletableFuture.completedFuture(offset);
```
gauravAshok, line 27: this impl assumes that we cannot do 

```
msgSrc.nextMessages(arr_1);
msgSrc.nextMessages(arr_2); // immediately 
```

which is fine, as that is how we are going to call it. but this needs to go into the documentation.

consumer/src/main/java/com/flipkart/varadhi/consumer/UnGroupedMessageSrc.java
```diff
@@ -0,0 +1,50 @@
...
+public class UnGroupedMessageSrc<O extends Offset> implements MessageSrc {
+
+    final Consumer<O> consumer;
+
+    private final BlockingQueue<MessageTracker> buffer = new LinkedBlockingQueue<>();
+
+    @Override
+    public CompletableFuture<Integer> nextMessages(MessageTracker[] messages) {
+        int offset = fetchFromBuffer(messages);
+        if (offset < messages.length) {
```
gauravAshok, line 23: Currently the impl is trying to fill the array completely. Though that is not strictly required, we should be able to return partial filled array.

consumer/src/main/java/com/flipkart/varadhi/consumer/PolledMessageTracker.java
```diff
@@ -0,0 +1,39 @@
...
+    @Override
+    public Message getMessage() {
+        return new PolledMessageWrapper<>(message);
+    }
+
+    @Override
+    public void onConsumed(MessageConsumptionStatus status) {
+        if (status == MessageConsumptionStatus.SENT || status == MessageConsumptionStatus.FILTERED) {
+            committer.commitIndividualAsync(message);
+        }
```
gauravAshok, line 27: you commit always :) It is not optional.

consumer/src/main/java/com/flipkart/varadhi/consumer/GroupedMessageSrc.java
```diff
@@ -0,0 +1,160 @@
...
+
+    private int nextMessagesInternal(MessageTracker[] messages) {
+        GroupTracker groupTracker = getGroupTracker();
+        if (null == groupTracker) {
+            return 0;
+        }
+
+        MessageTracker messageTracker = groupTracker.messages.getFirst().nextMessage();
+        messages[0] = new GroupedMessageTracker(messageTracker);
+        return 1;
```
gauravAshok, line 46: we are always returning 1 message?

We should be return many messages all from a unique group.

consumer/src/main/java/com/flipkart/varadhi/consumer/UnGroupedMessageSrc.java
```diff
@@ -0,0 +1,50 @@
...
+import java.util.concurrent.BlockingQueue;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.LinkedBlockingQueue;
+
+@RequiredArgsConstructor
+public class UnGroupedMessageSrc<O extends Offset> implements MessageSrc {
+
+    final Consumer<O> consumer;
+
+    private final BlockingQueue<MessageTracker> buffer = new LinkedBlockingQueue<>();
```
kmrdhruv, line 18: Does this need to be blocking given the current implementation and caller semantics ?

gauravAshok: I dont think so. Actually, it probably can be a simple arraylist protected by a lock. synchronziedList

kmrdhruv (changes requested):

consumer/src/main/java/com/flipkart/varadhi/consumer/UnGroupedMessageSrc.java
kmrdhruv: single threaded sequential execution is the expectation i.e. multiple nextMessages() concurrent execution is not expected, is synchronization still needed (since buffer is getting manipulated only in nextMessages())

gauravAshok: `since buffer is getting manipulated only in nextMessages` - not true, when msgSrc does not have any messages, then it will ask the consumer object for more which is async operation, so the new batch of messages will come on a different thread. From that thread, you will get the chance to populate this buffer.

consumer/src/main/java/com/flipkart/varadhi/consumer/MessageBatch.java
```diff
@@ -0,0 +1,33 @@
+package com.flipkart.varadhi.consumer;
+
+import java.util.List;
+
+public class MessageBatch {
```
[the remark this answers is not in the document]
aayustark007-fk: since we get messages in batches from the consumer, its better to just append those batches rather then doing the extra copy to move everything to single batch.

consumer/src/main/java/com/flipkart/varadhi/consumer/UnGroupedMessageSrc.java
aayustark007-fk: Isn't the ArrayList/Linked BlockingQueue roughly the same as list with locks?
Internal implementation is based on Michael and Scott's 2 lock concurrent queue which is designed to be more safe and performant.

Since, at most 2 threads will be accessing it, we can switch to the non-blocking ConcurrentLinkedQueue which is also based on the same paper. (2-lock is more suitable for cases where several threads are contending for shared resource) https://www.cs.rochester.edu/~scott/papers/1996_PODC_queues.pdf

consumer/src/main/java/com/flipkart/varadhi/consumer/UnGroupedMessageSrc.java
aayustark007-fk: I think the above scenario will cause two concurrent threads to read from the buffer (which will be thread safe) followed by batchReceiveAsync on the consumer.

From the pulsar docs, they recommend that this should be called subsequently only when prev completable future completes otherwise it can create a backlog of receive requests. 

So this assumption is mostly due to the concerns from the consumer side.

consumer/src/main/java/com/flipkart/varadhi/consumer/UnGroupedMessageSrc.java
aayustark007-fk: we can augment the consumer receive async future with `.orTimeout` so that in case no messages are available from the consumer we can still return early rather than waiting for a long time for messages.

consumer/src/main/java/com/flipkart/varadhi/consumer/UnGroupedMessageSrc.java
aayustark007-fk: PS: I just read this paper few weeks ago, was quite fascinated to find it's mention in the java doc for `ConcurrentLinkedQueue` and `BlockingQueue` implementation. These algorithms are quite ubiquitous!

consumer/src/main/java/com/flipkart/varadhi/consumer/PolledMessageTracker.java
aayustark007-fk: done

consumer/src/main/java/com/flipkart/varadhi/consumer/GroupedMessageSrc.java
aayustark007-fk: done

consumer/src/main/java/com/flipkart/varadhi/consumer/GroupedMessageSrc.java
```diff
@@ -0,0 +1,160 @@
...
+                    tracker.status = GroupStatus.FREE;
+                    isRemaining.setTrue();
+                    return tracker;
+                } else {
+                    return null;
+                }
+            });
+            totalInFlightMessages.decrementAndGet();
+            if (isRemaining.isTrue()) {
+                freeGroups.addFirst(groupId);
```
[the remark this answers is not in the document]
aayustark007-fk: I based this logic off the internal impl. So, not aware why we do that. I think @gauravAshok can explain in more detail.

consumer/src/main/java/com/flipkart/varadhi/consumer/UnGroupedMessageSrc.java
gauravAshok: So, these impl are meant to scale array operations when contention due to access from many threads is the issue. Is contention the issue in our case?

I am sure, array with a lock will be more performant in our case.

Actually, I think we dont even need array. Why can't  I just store the iterator reference from the Messages object that I am getting from consumer object?

consumer/src/main/java/com/flipkart/varadhi/consumer/UnGroupedMessageSrc.java
gauravAshok: isn't that redundant, when you can return even early knowing that you had to call consumer for its (time taking) async operation?

consumer/src/main/java/com/flipkart/varadhi/consumer/UnGroupedMessageSrc.java
aayustark007-fk: for iterator, we will need a lock when assigning it otherwise it can get overwritten in case due to a race condition.

gauravAshok: Hmm. I think this can be done without any locks. see if the usage of field `ConcurrencyControlImpl::schedulePendingTaskCounter` inspires you.

consumer/src/main/java/com/flipkart/varadhi/consumer/UnGroupedMessageSrc.java
aayustark007-fk: I think yes we can revert back to the previous behaviour where we return early once we fetch from the buffer

consumer/src/main/java/com/flipkart/varadhi/consumer/UnGroupedMessageSrc.java
aayustark007-fk: sure, I'll check that out

consumer/src/main/java/com/flipkart/varadhi/consumer/GroupedMessageSrc.java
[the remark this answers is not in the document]
gauravAshok: `wouldn't round robin across all free groups make more sense` - so, the freeGroup queue "roughly" represents the order of "groups" that we have encountered. If "g1" group is present before "g2" then it is likely that "g1" will be present before in the queue as well.

So, to unblock the "batch" and their commit early, we are, in a way, prioritizing the groups. Keep processing the groups that are present in the beginning of the queue. Hence we pop from head and if there are more to process for that group, we again put it back in the head so that it can be prioritized.

consumer/src/main/java/com/flipkart/varadhi/consumer/UnGroupedMessageSrc.java
AayuStark007: this is done

consumer/src/main/java/com/flipkart/varadhi/consumer/GroupedMessageSrc.java
```diff
@@ -0,0 +1,178 @@
...
+     * Attempt to fill the message array with one message from each group.
+     * Subsequent messages from a group are not fetched until the previous message is consumed.
+     *
+     * @param messages Array of message trackers to populate.
+     *
+     * @return CompletableFuture that completes when the messages are fetched.
+     */
+    @Override
+    public CompletableFuture<Integer> nextMessages(MessageTracker[] messages) {
+        if (hasMaxInFlightMessages()) {
```
gauravAshok, line 50: i think you meant (not hasMaxInFlightMessages)?

consumer/src/main/java/com/flipkart/varadhi/consumer/GroupedMessageSrc.java
```diff
@@ -0,0 +1,178 @@
...
+            return replenishAvailableGroups().thenApply(v -> nextMessagesInternal(messages));
+        }
+        return CompletableFuture.supplyAsync(() -> nextMessagesInternal(messages));
```
gauravAshok, line 53: It does not have to be CompletableFuture.supplyAsync

You can use CompletableFuture.completedFuture(), to imediately return with result. this supplyAsync introduces a unnecessary jump to another thread.

consumer/src/main/java/com/flipkart/varadhi/consumer/UnGroupedMessageSrc.java
```diff
@@ -0,0 +1,88 @@
...
+
+    private int processPolledMessages(int offset, PolledMessages<O> polledMessages, MessageTracker[] messages) {
+        int i = offset;
+
+        Iterator<PolledMessage<O>> polledMessagesIterator = polledMessages.iterator();
+        while (polledMessagesIterator.hasNext() && i < messages.length) {
+            PolledMessage<O> polledMessage = polledMessagesIterator.next();
+            MessageTracker messageTracker = new PolledMessageTracker<>(consumer, polledMessage);
+            messages[i++] = messageTracker;
+        }
```
gauravAshok, line 67: code duplication.
the method `fetchFromIterator` can take the iterator as parameter. so the main method can just use ongoingIterator as the param, and in this method, you can use polledMessagesIterator as the param.

PR #137, opened 2024-05-09
[DRAFT] Refactor Message class as an interface.  Model the producer message and polled message as implementations

aayustark007-fk:
This draft proposes the following changes:
- Refactor the Message class to Message interface
- For the concrete implementation of Message, introduce the ProducerMessage class which has the same functionality as the old Message class.
- For the consume side, messages can have an offset associated with them. Introduce the MessageWithOffset interface which extends the Message to add offset information.
- PolledMessage now extends the MessageWithOffset interface and by extension Message.

Benefit of the above is that the concept of Message is unified in the codebase and extensions are provided via PolledMessage and MessageWithOffset to model the consume side.
This also simplifies the MessageSrc implementation where we can implement MessageTracker over PolledMessage without writing the code to convert underlying PolledMessage to Message (it's simple now since the both share the same type)


entities/src/main/java/com/flipkart/varadhi/entities/MessageWithOffset.java
```diff
@@ -0,0 +1,9 @@
+package com.flipkart.varadhi.entities;
+
+public interface MessageWithOffset<O extends Offset> extends Message {
```
kmrdhruv, line 3: Does it make sense to merge it with PolledMessage, when it is needed independently meaning if there is Offset associated with message it should have other properties like topic/partition etc ?

pulsar/src/main/java/com/flipkart/varadhi/pulsar/entities/PulsarMessage.java
```diff
@@ -25,11 +28,41 @@ public PulsarOffset getOffset() {
         return new PulsarOffset(msg.getMessageId());
     }
 
+    @Override
+    public String getMessageId() {
+        throw new UnsupportedOperationException("Not implemented");
```
kmrdhruv, line 33: notimplementedexception ?

entities/src/main/java/com/flipkart/varadhi/entities/ProducerMessage.java
```diff
@@ -0,0 +1,56 @@
+package com.flipkart.varadhi.entities;
+
+import com.google.common.collect.ArrayListMultimap;
+import com.google.common.collect.Multimap;
+
+import java.util.List;
+
+public class ProducerMessage implements Message {
```
kmrdhruv, line 8: should this be moved to messaging.produce (PulsarProducer test might require a test fixture in entities for this)?

kmrdhruv (changes requested):

entities/src/main/java/com/flipkart/varadhi/entities/ProducerMessage.java
aayustark007-fk: PulsarProducerTest is able to access producer message via the entities dependency

entities/src/main/java/com/flipkart/varadhi/entities/MessageWithOffset.java
aayustark007-fk: Agree; we don't need a separate message with offset.
PolledMessage acts as the Message contract from the consumer side. It has the info on partitions and offsets.

pulsar/src/main/java/com/flipkart/varadhi/pulsar/entities/PulsarMessage.java
aayustark007-fk: done

consumer/src/main/java/com/flipkart/varadhi/consumer/PolledMessageTracker.java
```diff
@@ -20,22 +19,11 @@ public PolledMessageTracker(Consumer<O> committer, PolledMessage<O> message) {
...
     public Message getMessage() {
-        return new PolledMessageWrapper<>(message);
+        return message;
     }
 
     @Override
     public void onConsumed(MessageConsumptionStatus status) {
         committer.commitIndividualAsync(message);
     }
-
```
aayustark007-fk, line 30: since PolledMessage is now a Message, we don't need to implement a Wrapper to make it work with the MessageTracker contracts

pulsar/src/main/java/com/flipkart/varadhi/pulsar/entities/PulsarMessage.java
```diff
@@ -25,11 +29,41 @@ public PulsarOffset getOffset() {
         return new PulsarOffset(msg.getMessageId());
     }
 
+    @Override
```
aayustark007-fk, line 32: need to figure out, how to conform to the Message contracts around headers, message and group Id

aayustark007-fk (commented):
.

entities/src/main/java/com/flipkart/varadhi/entities/ProducerMessage.java
kmrdhruv: ok, suggestion was to check if this can be moved to messaging.produce as this is specific to produce path alone. ?

pulsar/src/main/java/com/flipkart/varadhi/pulsar/entities/PulsarMessage.java
kmrdhruv: planning to take it independently ?

kmrdhruv (approved):
minor comments, please check prior to merge.

PR #145, opened 2024-05-24
making fixes to grouped message src, so that nextMessages always retu…

gauravAshok:
…rns with messages asynchronously. previously it could return with 0 msgs and used to prioritize fetching messages from consumer

### DIFF
--- a/consumer/src/main/java/com/flipkart/varadhi/consumer/GroupedMessageSrc.java
+++ b/consumer/src/main/java/com/flipkart/varadhi/consumer/GroupedMessageSrc.java
@@ -1,5 +1,6 @@
 package com.flipkart.varadhi.consumer;
 
+import com.flipkart.varadhi.consumer.concurrent.Context;
 import com.flipkart.varadhi.entities.Message;
 import com.flipkart.varadhi.entities.Offset;
 import com.flipkart.varadhi.spi.services.Consumer;
@@ -7,6 +8,7 @@ import com.flipkart.varadhi.spi.services.PolledMessage;
 import com.flipkart.varadhi.spi.services.PolledMessages;
 import lombok.AllArgsConstructor;
 import lombok.RequiredArgsConstructor;
+import lombok.extern.slf4j.Slf4j;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.commons.lang3.mutable.MutableBoolean;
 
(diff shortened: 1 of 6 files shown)


consumer/src/main/java/com/flipkart/varadhi/consumer/GroupedMessageSrc.java
```diff
@@ -158,7 +214,7 @@ private void free(String groupId, MessageConsumptionStatus status) {
                     throw new IllegalStateException(String.format("Tried to free group %s: %s", gId, tracker));
                 }
                 var messages = tracker.messages;
-                if (!messages.isEmpty() && messages.getFirst().remaining() == 0) {
```
gauravAshok, line 161: using `while` to make sure that messages batch with empty entries are not going to be put back in the hashmap.

consumer/src/main/java/com/flipkart/varadhi/consumer/GroupedMessageSrc.java
```diff
@@ -47,22 +66,65 @@ public class GroupedMessageSrc<O extends Offset> implements MessageSrc {
      */
     @Override
     public CompletableFuture<Integer> nextMessages(MessageTracker[] messages) {
-        if (!hasMaxInFlightMessages()) {
-            return replenishAvailableGroups().thenApply(v -> nextMessagesInternal(messages));
+        int count = nextMessagesInternal(messages);
+        if (count > 0) {
+            return CompletableFuture.completedFuture(count);
```
gauravAshok, line 71: prioritizing the array filling.

If no message present / no free group present. try to fetch fresh messages.

And this is the difference, There are 2 conditions, which can unblock the "array filling". the fetch returning with new messages.
and the group getting freed up.

consumer/src/main/java/com/flipkart/varadhi/consumer/GroupedMessageSrc.java
```diff
@@ -47,22 +66,65 @@ public class GroupedMessageSrc<O extends Offset> implements MessageSrc {
...
+            tryCompletePendingRequest();
+        }
+
+        return request.result;
+    }
+
+    private void tryCompletePendingRequest() {
+        NextMsgsRequest request;
+        if ((request = pendingRequest.getAndSet(null)) != null) {
+            request.result.complete(nextMessagesInternal(request.messages));
```
gauravAshok, line 94: there is still a chance, where we still end up completing the request with 0 messages, but this should be much rarer now.

aayustark007-fk (approved):

PR #152, opened 2024-06-07
Implement missing methods in PulsarMessage

aayustark007-fk:
This PR implements the missing methods in PulsarMessage which had dependency on the request headers which were passed in the Produce path.

Since, the produce path added the request headers as Pulsar API Message properties, all we need to do is to fetch the properties and extract all the key value pairs to rebuild the request headers.

Now we can implement the missing methods like getHeader(), getMessageId, getGroupId, etc


kmrdhruv (approved):
LGTM

PR #158, opened 2024-07-22
Retry for failed operation

kmrdhruv:
Initial implementation of retry failed operation in controller.  Retry is based on simple retry policy and will always be attempted. Currently retry is failure agnostic.

TODO:: working on tests.


controller/src/main/java/com/flipkart/varadhi/controller/RetryPolicy.java
```diff
@@ -0,0 +1,23 @@
+package com.flipkart.varadhi.controller;
+
+import com.flipkart.varadhi.entities.cluster.OrderedOperation;
+import lombok.AllArgsConstructor;
+
+@AllArgsConstructor
+public class RetryPolicy {
```
AayuStark007, line 7: can use builder with defaults here to allow easy construction of the instance

entities/src/main/java/com/flipkart/varadhi/entities/cluster/ShardStatus.java
```diff
@@ -21,4 +21,15 @@ public boolean canStart() {
         return state == ShardState.STOPPED || state == ShardState.ERRORED || state == ShardState.UNKNOWN ||
                 state == ShardState.STOPPING;
     }
+
+    @JsonIgnore
+    public boolean isStarted() {
+        return state == ShardState.STARTED ||  state == ShardState.STARTING;
```
AayuStark007, line 27: state should be final and non-null checked, else it can lead to npe

AayuStark007 (changes requested):

entities/src/main/java/com/flipkart/varadhi/entities/cluster/ShardStatus.java
kmrdhruv: done

controller/src/main/java/com/flipkart/varadhi/controller/RetryPolicy.java
kmrdhruv: Not doing this for now as not sure of final policy struct, also policy is created only once in code (excluding test). Would prefer to avoid test defaults in prod code.

PR #165, opened 2024-09-16
Use address for remote calls in vertx clustered event bus

iosdev747:
PR changes:
- HostUtils has to be init before using it in the app/tests. (On init it stores the hostname and address or throws exceptions to fail the startup).
- To address running app in k8s env, vertx clustered event bus will be using IP address for public comms (over the wire) and hostname for node identification.

### DIFF
--- a/common/src/main/java/com/flipkart/varadhi/utils/HostUtils.java
+++ b/common/src/main/java/com/flipkart/varadhi/utils/HostUtils.java
@@ -8,11 +8,18 @@ import java.net.UnknownHostException;
 @Slf4j
 public class HostUtils {
 
-    public static String getHostName() throws UnknownHostException {
-        // debug to see how much time it takes, in case DNS resolution is taking time.
-        log.debug("getHostName: started");
-        String host = InetAddress.getLocalHost().getHostName();
-        log.debug("getHostName: completed");
-        return host;
+    public static String getHostNameOrAddress(boolean requireHostName) throws UnknownHostException {
+        if(requireHostName) {
+            // debug to see how much time it takes, in case DNS resolution is taking time.
+            log.debug("getHostName: started");
+            String host = InetAddress.getLocalHost().getHostName();
+            log.debug("getHostName: completed");
+            return host;
+        } else {
+            log.debug("getAddress: started");
+            String address = InetAddress.getLocalHost().getHostAddress();
+            log.debug("getAddress: completed");
+            return address;
+        }
     }
 }
--- a/common/src/main/java/com/flipkart/varadhi/utils/HostUtils.java
+++ b/common/src/main/java/com/flipkart/varadhi/utils/HostUtils.java
@@ -15,4 +15,12 @@ public class HostUtils {
         log.debug("getHostName: completed");
         return host;
     }
+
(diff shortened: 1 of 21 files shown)


server/src/main/java/com/flipkart/varadhi/VaradhiApplication.java
```diff
@@ -95,7 +95,7 @@ private static Future<Vertx> createClusteredVertx(
         int port = 0;
         JsonObject memberInfoJson = new JsonObject(JsonMapper.jsonSerialize(memberInfo));
         EventBusOptions eventBusOptions = new EventBusOptions()
-                .setHost(memberInfo.hostname())
+                .setHost(memberInfo.host())
```
kmrdhruv, line 98: EventBus has an option to set clusterpublichost. May be we can use host to be the podname and use ip to set publichost. Need to see node will get registered (via ip or name) in that case ?

kmrdhruv (changes requested):

server/src/main/java/com/flipkart/varadhi/VaradhiApplication.java
iosdev747: made the change. Now relying on clusterpublichost.

common/src/main/java/com/flipkart/varadhi/utils/HostUtils.java
```diff
@@ -15,4 +15,12 @@ public static String getHostName() throws UnknownHostException {
         log.debug("getHostName: completed");
         return host;
     }
+
+    public static String getHostAddress() throws UnknownHostException {
```
gauravAshok, line 19: dont log. dont throw

gauravAshok: faced the same issue internally. where calling this function unnecesarilly requires exception handling.

call this function at the app startup and if it throws exception fail the startup itself. If it works once, I think it will work always, hence there is no need to throw checked exception.

server/src/main/java/com/flipkart/varadhi/VaradhiApplication.java
```diff
@@ -97,6 +98,7 @@ private static Future<Vertx> createClusteredVertx(
         EventBusOptions eventBusOptions = new EventBusOptions()
                 .setHost(memberInfo.hostname())
                 .setPort(port)
+                .setClusterPublicHost(memberInfo.address())
```
gauravAshok, line 101: in eventbus options, what is host used for if cluster public host is already present?

common/src/main/java/com/flipkart/varadhi/utils/HostUtils.java
iosdev747: Done, now at startup HostUtils init is required. Post that get calls doesn't need to handle exceptions.

server/src/main/java/com/flipkart/varadhi/VaradhiApplication.java
iosdev747: host will be used for nodeId (internal to vertx) but for actual comms, address will be used.

kmrdhruv (approved):
LGTM

PR #166, opened 2024-10-01
otel collector for metric collection.

kmrdhruv:
Works for dev setup for metric collection.

For prod additional changes will be needed.
- Convert the otel container as a sidecar
- Test traces and logs in addition to metrics.
- Promethous configuration for scrapping needs to be defined.


gaurav-ashok (approved):
LGTM

PR #198, opened 2024-12-09
DLQ getMessages() API refactoring

kmrdhruv:
- abstracted the offsets
- converted fro post to get.

pending testsi,


spi/src/main/java/com/flipkart/varadhi/spi/services/PolledMessage.java
```diff
@@ -24,12 +24,6 @@ public interface PolledMessage<O extends Offset> extends Message {
      */
     O getOffset();
 
-    /**
-     * @return the payload of this message.
-     */
-    // TODO: evaluate ByteBuf instead of byte[]. byte[] means heap allocated.
-    byte[] getPayload();
```
gauravAshok, line 31: why did we remove this method?

entities/src/main/java/com/flipkart/varadhi/entities/TopicPartitions.java
```diff
@@ -20,7 +20,9 @@ public static <T extends StorageTopic> TopicPartitions<T> byPartitions(T topic,
     }
 
     public static <T extends StorageTopic> TopicPartitions<T> byTopic(T topic) {
-        return new TopicPartitions<>(topic, null);
+        //TODO:: Discuss this
+        // Can a topic be non partitioned or assume there is always 1 (implicit).
```
gauravAshok, line 24: yes, the topic can be "non-partitioned", or even if partitioned, then using this method means implicitly that you are talking about the topic as a whole (including all of its partitions).

core/src/main/java/com/flipkart/varadhi/core/cluster/ConsumerApi.java
```diff
@@ -21,5 +19,7 @@ public interface ConsumerApi {
 
     CompletableFuture<ConsumerInfo> getConsumerInfo();
 
-    CompletableFuture<List<Message>> getMessages(GetMessagesRequest messagesRequest);
+    CompletableFuture<ShardDlqMessageResponse> getMessagesByTimestamp(long earliestFailedAt, int max_limit);
+
+    CompletableFuture<ShardDlqMessageResponse> getMessagesByOffset(String pageMarkers, int max_limit);
```
gauravAshok, line 24: String pageMarkers: "String" type looks too generic.
Why is it a string and how this string should be interpreted?

gauravAshok: ok. got it. the pageMarkers are "abstract" even for the web-api. These are returned by the consumers as part of the ShardDLQMessageResponse.

So, consumer come up with this value, and they can understand them. for the other components, it is just a abstract string.

core/src/main/java/com/flipkart/varadhi/core/cluster/entities/ShardDlqMessageRequest.java
```diff
@@ -0,0 +1,30 @@
...
+
+import java.util.ArrayList;
+import java.util.List;
+
+@Data
+@AllArgsConstructor(onConstructor = @__(@JsonCreator))
+public class ShardDlqMessageRequest {
+    private long earliestFailedAt = 0L;
+    private final String pageMarker;
+    private int limit = -1;
```
gauravAshok, line 17: what would limit = -1 mean?

server/src/main/java/com/flipkart/varadhi/web/entities/DlqPageMarker.java
```diff
@@ -0,0 +1,70 @@
...
+import java.util.List;
+import java.util.Map;
+import java.util.regex.Pattern;
+import java.util.stream.Collectors;
+
+
+@Data
+@Slf4j
+public class DlqPageMarker {
+    private static final Pattern validPargeMarkerPattern = Pattern.compile("\\d+=([^~]+)(?:[~]\\d+=([^~]+))*[~]?");
```
gauravAshok, line 17: typo

server/src/main/java/com/flipkart/varadhi/services/DlqService.java
```diff
@@ -34,37 +42,55 @@ public CompletableFuture<SubscriptionOperation> unsideline(
...
-                                            "Failed to getMessages for {}(shard:{}). Error:{}", subscription.getName(),
-                                            assignment.getShardId(), t.getMessage()
-                                    );
-                                    shardResponse.setError(t.getMessage());
-                                }
-                                log.info("Sending a GetMessage response for {}", assignment);
-                                responseWriter.accept(shardResponse);
-                            });
-                        }).toArray(CompletableFuture[]::new)));
+        ShardDlqMsgResponseCollector finalResponse = new ShardDlqMsgResponseCollector();
```
gauravAshok, line 57: This file is pending for review, everything else is reviewed.

core/src/main/java/com/flipkart/varadhi/core/cluster/ConsumerApi.java
kmrdhruv: yes, that's the intent.

spi/src/main/java/com/flipkart/varadhi/spi/services/PolledMessage.java
kmrdhruv: its available in Message interface which is being extended here.

server/src/main/java/com/flipkart/varadhi/web/entities/DlqPageMarker.java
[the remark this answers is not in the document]
kmrdhruv: done

core/src/main/java/com/flipkart/varadhi/core/cluster/entities/ShardDlqMessageRequest.java
```diff
@@ -0,0 +1,30 @@
...
+
+    public ShardDlqMessageRequest(long earliestFailedAt, int limit) {
+        this.earliestFailedAt = earliestFailedAt;
+        this.pageMarker = null;
+        this.limit = limit;
+    }
+
+    public ShardDlqMessageRequest(String pageMarker, int limit) {
+        earliestFailedAt = 0L;
```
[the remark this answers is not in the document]
kmrdhruv: Not sure, i guess earliest and latest would translate to LONG_MIN and LONG_MAX semantics and hence used 0 as another sentinel to indicate unspecified.

core/src/main/java/com/flipkart/varadhi/core/cluster/entities/ShardDlqMessageRequest.java
kmrdhruv: that was to indicate default value, but not a valid scenraio.
removed default initialization.

server/src/main/java/com/flipkart/varadhi/services/DlqService.java
```diff
@@ -34,37 +42,55 @@ public CompletableFuture<SubscriptionOperation> unsideline(
...
+        boolean isRequestByTimeStamp = earliestFailedAt != UNSPECIFIED_TS;
+        return controllerClient.getShardAssignments(subscription.getName()).thenCompose(assignments -> {
+            List<CompletableFuture<ShardDlqMessageResponse>> shardFutures = new ArrayList<>();
+            for (Assignment a : assignments.getAssignments()) {
+                int shardId = a.getShardId();
+                if (!isRequestByTimeStamp && !pageMarkers.hasMarker(shardId)) {
+                    log.info("Shard {} has no markers, skipping getMessages().", shardId);
+                    continue;
+                }
```
gauravAshok, line 66: We are skipping a shard if info not provided. Is this intentional?

Should we just fail the request outright as bad request if a marker is missing for any shard?

server/src/main/java/com/flipkart/varadhi/services/DlqService.java
```diff
@@ -34,37 +42,55 @@ public CompletableFuture<SubscriptionOperation> unsideline(
...
+            ShardDlqMsgResponseCollector finalResponse
+    ) {
+        if (r != null && !r.getMessages().isEmpty()) {
+            log.info(
+                    "shard {} returned {} messages nextMarker {}.", shardId, r.getMessages().size(),
+                    r.getNextPageMarker()
+            );
+            responseWriter.accept(DlqMessagesResponse.of(r.getMessages()));
+        }
+        finalResponse.collectShardResponse(shardId, t, r);
```
gauravAshok, line 86: so it looks like, in the response, we are writing an array?
having "shard + 1" elements. where the first "shard" elements are the messages per shard and the last element is the info about the pageMarkers and errors.

Is this understanding correct?

entities/src/main/java/com/flipkart/varadhi/entities/TopicPartitions.java
```diff
@@ -20,7 +20,7 @@ public static <T extends StorageTopic> TopicPartitions<T> byPartitions(T topic,
     }
 
     public static <T extends StorageTopic> TopicPartitions<T> byTopic(T topic) {
-        return new TopicPartitions<>(topic, null);
+        return new TopicPartitions<>(topic,null);
```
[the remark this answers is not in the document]
kmrdhruv: fixed

server/src/main/java/com/flipkart/varadhi/services/DlqService.java
kmrdhruv: yes, 1 response per shard and followed by final response with all marker/errors.

server/src/main/java/com/flipkart/varadhi/services/DlqService.java
gauravAshok: I would guess,  this may be helpful in fetching messages from a partial set of shards.

server/src/main/java/com/flipkart/varadhi/services/DlqService.java
gauravAshok: Around the chunked response. The overall repsonse is  not a valid json
so that is an issue.

I am expecting that overall, the response headers will contain `content-type : application/json`.
which will cause the query on UI to break, as that will try to parse the chunked response as a full json

server/src/main/java/com/flipkart/varadhi/services/DlqService.java
kmrdhruv: yes. If marker is not available, it is assumed there are no further messages to be returned and hence such shards are skipped. It is also assumed users are not manipulating the markers as they are abstracted from them.

PR #217, opened 2025-01-27
addding interface for AuthenticationHandlerProvider

alekhrj:
## Summary by CodeRabbit

- **New Features**
  - Enhanced authentication system with new interfaces and configuration options.
  - Added support for user context and authentication provider.
  - Introduced methods to retrieve user attributes.
  - New class `AuthenticationOptions` to manage authentication configurations.
  - New class `RequestContext` for structured access to request components.

- **Improvements**
  - Expanded module exports to support new authentication-related packages.
  - Implemented flexible authentication handling with asynchronous methods.

### DIFF
--- a/entities/src/main/java/com/flipkart/varadhi/entities/auth/UserContext.java
+++ b/entities/src/main/java/com/flipkart/varadhi/entities/auth/UserContext.java
@@ -1,6 +1,7 @@
 package com.flipkart.varadhi.entities.auth;
 
 import java.util.Map;
+import java.util.Optional;
 
 public interface UserContext {
 
@@ -8,9 +9,13 @@ public interface UserContext {
 
     boolean isExpired();
 
-    String getOrg();
+    default Optional<String> getOrg() {
+        return Optional.empty();
+    }
 
-    Map<String, String> getAttributes();
+    default Map<String, String> getAttributes() {
+        return Map.of();
+    }
 
     // TODO: enhance it to include more details like context around the authentication
 }
--- a/entities/src/main/java/com/flipkart/varadhi/entities/auth/UserContext.java
+++ b/entities/src/main/java/com/flipkart/varadhi/entities/auth/UserContext.java
@@ -1,7 +1,6 @@
 package com.flipkart.varadhi.entities.auth;
 
 import java.util.Map;
-import java.util.Optional;
 
 /*
     * UserContext represents the authenticated user. It is used to pass the user information across the system.
@@ -13,10 +12,6 @@ public interface UserContext {
     String getSubject();
 
     boolean isExpired();
-
-    default Optional<String> getOrg() {
-        return Optional.empty();
-    }
 
     default Map<String, String> getAttributes() {
         return Map.of();
--- a/entities/src/main/java/com/flipkart/varadhi/entities/auth/UserContext.java
+++ b/entities/src/main/java/com/flipkart/varadhi/entities/auth/UserContext.java
@@ -3,6 +3,11 @@ package com.flipkart.varadhi.entities.auth;
 import java.util.Map;
 import java.util.Optional;
 
+/*
+    * UserContext represents the authenticated user. It is used to pass the user information across the system.
+    * This can be enhanced to include more details like context around the authentication.
+*/
+
 public interface UserContext {
 
     String getSubject();
@@ -16,6 +21,4 @@ public interface UserContext {
     default Map<String, String> getAttributes() {
         return Map.of();
     }
-
-    // TODO: enhance it to include more details like context around the authentication
 }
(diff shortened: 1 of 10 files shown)


spi/src/main/java/com/flipkart/varadhi/spi/authn/AuthenticationHandlerProvider.java
```diff
@@ -0,0 +1,8 @@
+package com.flipkart.varadhi.spi.authn;
+
+import io.vertx.core.Vertx;
+import io.vertx.ext.web.handler.AuthenticationHandler;
+
+public interface AuthenticationHandlerProvider {
+    AuthenticationHandler provideHandler(Vertx vertx, String fileName);
```
gauravAshok, line 7: why this method signature specifically?

which fileName is this?

Why is vertx instance required?

gauravAshok: Also, we now have exposed vertx as the dependency in our spi. Is that correct direction?

kmrdhruv: yeah.. better to avoid any specific framework here.
I think earlier also AuthorizationProvider is also using Vertx for its Future. We can relook at it as well.

alekhrj: 1. fileName is config file name for configuring specific authentication handler/provider. will update the variable name. 
2. vertx instance is needed to create AuthenticationProviders, which are embedded in the AuthenticationHandlers for JWT/Oauth2 providers as mentioned here - https://vertx.io/docs/vertx-auth-oauth2/java/

gauravAshok: can't the provider create their own vertx instance?

AuthenticationHandler is just a Handler<RoutingContext>, which is very http centric. Any AuthenticationHandler implementation should come OOB and should be configurable.

Anything else, we need to strip it down. like our own 
VRDAuthenticationHandler::handle(URI, params, headers, Map<String, Object> context) -> Future<AuthenticationResult>

this stripped down interface can be part of spi.

Just a thought.

entities/src/main/java/com/flipkart/varadhi/entities/auth/UserContext.java
```diff
@@ -1,10 +1,24 @@
...
+*/
+
 public interface UserContext {
 
     String getSubject();
 
     boolean isExpired();
 
-    // TODO: enhance it to include more details like context around the authentication
+    default Optional<String> getOrg() {
```
gauravAshok, line 17: what does it mean for the user to have Org?

spi/src/main/java/com/flipkart/varadhi/spi/authn/AuthenticationProvider.java
```diff
@@ -0,0 +1,13 @@
...
+
+import com.flipkart.varadhi.entities.auth.UserContext;
+import com.flipkart.varadhi.spi.ConfigFileResolver;
+import io.vertx.core.Future;
+
+import java.net.URI;
+import java.util.Map;
+
+public interface AuthenticationProvider {
+    Future<Boolean> init(ConfigFileResolver resolver, AuthenticationOptions authenticationOptions);
```
gauravAshok, line 11: for now I think we can omit the resolver parameter. Why is it needed?

alekhrj: yes. this can be removed.

entities/src/main/java/com/flipkart/varadhi/entities/auth/UserContext.java
alekhrj: org here is associated with the tenants of AuthN. org/tenants have been used interchangeably by authN. 

for varadhi also, cross-org resource access should not be allowed.

entities/src/main/java/com/flipkart/varadhi/entities/utils/RequestContext.java
```diff
@@ -0,0 +1,14 @@
...
+
+import java.net.URI;
+import java.util.Map;
+
+@Data
+public class RequestContext {
+    private URI getURI;
+    private Map<String, String> params;
+    private Map<String, String> headers;
+    private Map<String, Object> gontext;
```
gauravAshok, line 13:
```suggestion
    private Map<String, Object> context;
```
🤣

entities/src/main/java/com/flipkart/varadhi/entities/utils/RequestContext.java
```diff
@@ -0,0 +1,14 @@
+package com.flipkart.varadhi.entities.utils;
+
+import lombok.Data;
+
+import java.net.URI;
+import java.util.Map;
+
+@Data
+public class RequestContext {
+    private URI getURI;
```
gauravAshok, line 10:
```suggestion
    private URI uri;
```

gauravAshok (approved):
LGTM

PR #244, opened 2025-03-05
#236 | Cache Synchronisation Handler and Warmup Mechanism

devanshu-garg-IN:
## Summary by CodeRabbit

- **New Features**
  - Introduced a generic, thread-safe entity cache system for efficient entity retrieval and event-driven updates.
  - Added a centralized registry to manage entity caches by resource type.
  - Implemented a type-safe entity event dispatcher for handling and routing cluster-wide entity change events.
  - Project and topic services now support full entity listing.

- **Refactor**
  - Replaced custom caching with a unified entity cache and registry across services and handlers.
  - Updated service and handler constructors to use the new cache interfaces.
  - Improved modularity and asynchronous control flow in core verticles and application startup.
  - Enhanced type safety in event listener interfaces and event processing.
  - Simplified producer service caching with Caffeine and unified topic retrieval.
  - Removed caching from project service and added detailed validation and logging.
  - Updated handlers and tests to depend on entity caches instead of services for project data.

- **Bug Fixes**
  - Improved error messages and validation in project and subscription operations.
  - Standardized error message formats in tests and handlers.

- **Chores**
  - Added new dependencies for caching and asynchronous testing.
  - Updated configuration files to reflect new cache settings.
  - Added Vert.x core dependency and updated module exports.
  - Added metrics and tracing improvements during clustered Vert.x initialization.

- **Tests**
  - Refactored tests to use the new entity cache system and updated mocking strategies.
  - Adjusted test expectations to match updated error messages and service behaviors.
  - Added delays in end-to-end tests to ensure cache consistency.

- **Documentation**
  - Added and improved Javadoc and inline documentation for new and refactored classes.

- **Style**
  - Reorganized imports and improved code formatting for readability.


common/src/main/java/com/flipkart/varadhi/common/VaradhiCache.java
```diff
@@ -1,83 +1,208 @@
...
  */
-public class VaradhiCache<K, V> {
-    // This is NonLoading cache. Get wll invoke loader() on calling thread when needed and hence possibly blocking in
-    // some cases.
-    // TODO:: need to support unblocking loader call.
-    private final Cache<K, V> entityCache;
-    private final Function<K, V> entityProvider;
-    private final Counter getCounter;
+@Slf4j
+public final class VaradhiCache<K, V> {
```
gauravAshok, line 23: lets just use https://github.com/ben-manes/caffeine?

core/src/main/java/com/flipkart/varadhi/core/cluster/EventListener.java
```diff
@@ -0,0 +1,24 @@
...
+import com.flipkart.varadhi.common.events.EntityEvent;
+
+/**
+ * Listener interface for entity events in the Varadhi system.
+ * <p>
+ * Implementations of this interface can process entity events for different resource types
+ * such as projects, topics, etc. The listener is responsible for handling state changes
+ * based on the events received.
+ */
+public interface EventListener {
```
gauravAshok, line 12: duplicate of EntityEventListener

server/src/main/java/com/flipkart/varadhi/events/ClusterEventDispatcher.java
```diff
@@ -0,0 +1,66 @@
...
+import java.util.concurrent.CompletableFuture;
+
+/**
+ * Dispatches entity events received via cluster messages to the appropriate event listeners.
+ * <p>
+ * This class acts as a bridge between the cluster messaging system and the entity event processing system,
+ * extracting entity events from cluster messages and dispatching them to the appropriate event listener.
+ */
+@Slf4j
+public final class ClusterEventDispatcher {
```
gauravAshok, line 19: this class looks redundant.

common/build.gradle
```diff
@@ -13,6 +13,7 @@ dependencies {
     implementation("com.fasterxml.jackson.datatype:jackson-datatype-jdk8:$jackson_version")
     implementation("io.micrometer:micrometer-core")
     implementation('org.hibernate:hibernate-validator:8.0.1.Final')
+    implementation("com.github.ben-manes.caffeine:caffeine:3.2.0")
```
gauravAshok, line 16: lets use the lib manifest way of declaring the dependency.

Check:
https://github.com/flipkart-incubator/varadhi/blob/83c829a91de673379e43613bdb6fa36af0e96981/server/build.gradle#L21
&
https://github.com/flipkart-incubator/varadhi/blob/83c829a91de673379e43613bdb6fa36af0e96981/gradle/libs.versions.toml#L13

common/src/main/java/com/flipkart/varadhi/common/VaradhiCache.java
gauravAshok: I think we can literally remove this class. I dont see a need for this class. If we need a cache then we can directly use caffiene

producer/src/main/java/com/flipkart/varadhi/produce/services/ProducerService.java
```diff
@@ -2,94 +2,161 @@
...
+
+    @Getter
     private final VaradhiCache<StorageTopic, Producer> producerCache;
+
+    @Getter
     private final VaradhiCache<String, VaradhiTopic> internalTopicCache;
+
     private final String produceRegion;
+    private final Function<StorageTopic, Producer> producerProvider;
+    private final MetaStore metaStore;
```
gauravAshok, line 50: not needed.

server/src/main/java/com/flipkart/varadhi/verticles/webserver/WebServerVerticle.java
```diff
@@ -289,30 +516,120 @@ private void setupRouteConfigurators() {
...
     }
+
+    /**
+     * Sets up event handlers for entity events.
+     *
+     * @return a Future that completes when event handlers are set up
+     */
+    @SuppressWarnings ("unchecked")
+    private Future<Void> setupEventHandlers() {
+        Promise<Void> promise = Promise.promise();
```
gauravAshok, line 569: When a app starts:
- connect all the dbs
- init vertx
- vertx cluster (inter node events will start)
- event processor needs to be init
- start the web server verticles.



move this initialization to probably in its own class.

server/src/main/java/com/flipkart/varadhi/events/EntityStateProcessor.java
```diff
@@ -0,0 +1,270 @@
...
+ * Processes entity state changes and updates in-memory state accordingly.
+ * <p>
+ * This processor uses a functional approach to manage entity state updates for different resource types,
+ * providing a clean and extensible way to handle entity events across the system.
+ * <p>
+ * The processor directly handles cluster messages containing entity events, extracting the events
+ * and processing them appropriately.
+ */
+@Slf4j
+public final class EntityStateProcessor {
```
gauravAshok, line 36: it is a dispatcher. it should only contain dispatch logic.

project, topic related code probbaly should go into their own service class.

server/src/main/java/com/flipkart/varadhi/services/ProjectService.java
```diff
@@ -1,24 +1,89 @@
...
+import java.util.stream.Collectors;
 
+/**
+ * Service for managing projects in Varadhi.
+ * <p>
+ * This service provides methods for creating, retrieving, updating, and deleting projects,
+ * as well as managing a cache of projects for efficient access.
+ */
+@Slf4j
 public class ProjectService {
```
gauravAshok, line 27: // retain the original ProjectService. CRUD from DB. No cache.
// create a diff class CacheProjectSupplier/Provider/Getter. get(name) -> Proejct. implement the entitiyEventListener

gauravAshok (changes requested):
public interface EntityEventListener\<T\> {

    /**
     * Called when an entity changes.
     * <p>
     * This method is invoked by event sources when an entity is created, updated, or deleted.
     * Implementations should process the event appropriately based on their specific requirements.
     *
     * @param event the event containing information about the entity change
     * @throws IllegalStateException if the listener is in a state where it cannot process events
     */
    void onChange(EntityEvent\<? extends T\> event);
}

server/src/main/java/com/flipkart/varadhi/services/ProjectService.java
gauravAshok: btw. CacheProjectSupplier / CachedTopicSupplier / etc.. all will have the same class. The pattern is to preload the data and rely on events.

so maybe a single class to abstract this makes sense. Then you can just do ReadCache\<Project\>, ReadCache\<Topic\>, etc.

common/src/main/java/com/flipkart/varadhi/common/EntityProvider.java
```diff
@@ -0,0 +1,165 @@
...
+ * Implementations must provide entity loading and naming strategies by implementing
+ * the abstract methods {@link #loadAllEntities()} and {@link #getEntityName(Object)}.
+ * <p>
+ * This class is optimized for concurrent access patterns with a focus on read-heavy workloads,
+ * using a {@link ConcurrentHashMap} for storage which provides thread-safety for individual operations.
+ *
+ * @param <T> the type of entity managed by this provider
+ */
+@Slf4j
+public abstract class EntityProvider<T> implements EntityEventListener<T> {
```
gauravAshok, line 37: lets rename it to EntityReadCache<T>

common/src/main/java/com/flipkart/varadhi/common/EntityProvider.java
```diff
@@ -0,0 +1,165 @@
...
+
+    /**
+     * Returns the name of the given entity.
+     * <p>
+     * Implementations must provide a concrete naming strategy.
+     *
+     * @param entity the entity to get the name for
+     * @return the name of the entity
+     */
+    protected abstract String getEntityName(T entity);
```
gauravAshok, line 86: isn't T here something that extends MetastoreEntity. which does have a name. I think this method is not needed.

common/src/main/java/com/flipkart/varadhi/common/EntityProvider.java
```diff
@@ -0,0 +1,165 @@
...
+     */
+    @Override
+    public void onChange(EntityEvent<T> event) {
+        String entityName = event.resourceName();
+        EventType operation = event.operation();
+
+        switch (operation) {
+            case UPSERT -> {
+                T entity = event.resource();
+                if (entity != null) {
```
gauravAshok, line 147: I dont think it can be null.

common/src/main/java/com/flipkart/varadhi/common/EntityProvider.java
```diff
@@ -0,0 +1,165 @@
...
+                } else {
+                    log.warn("Received UPSERT event with null resource for {}: {}", entityTypeName, entityName);
+                }
+            }
+            case INVALIDATE -> {
+                T removed = entities.remove(entityName);
+                if (removed != null) {
+                    log.debug("Removed {} from local cache: {}", entityTypeName, entityName);
+                } else {
+                    log.debug("Attempted to remove non-existent {} from cache: {}", entityTypeName, entityName);
```
gauravAshok, line 159: can be warn

common/src/main/java/com/flipkart/varadhi/common/EntityProvider.java
```diff
@@ -0,0 +1,165 @@
...
+    protected final String entityTypeName;
+
+    /**
+     * Creates a new entity provider for the specified resource type.
+     *
+     * @param resourceType   the type of resource managed by this provider
+     * @param entityTypeName a human-readable name for the entity type
+     * @throws NullPointerException if resourceType or entityTypeName is null
+     */
+    protected EntityProvider(ResourceType resourceType, String entityTypeName) {
```
gauravAshok, line 62: why can't we use the resourceType.toString() as the entityTypeName?

common/src/main/java/com/flipkart/varadhi/common/EntityProvider.java
```diff
@@ -0,0 +1,165 @@
...
+    }
+
+    /**
+     * Loads all entities from the underlying data source.
+     * <p>
+     * Implementations must provide a concrete loading strategy.
+     *
+     * @return a list of all entities
+     */
+    protected abstract List<T> loadAllEntities();
```
gauravAshok, line 76: I think instead of abstract class we can go with strategy here.
You just need a preloader method.

new EntityReadCache<T>(ResourceType resourceType, Supplier<List<T>> preloader);

`TopicReadCache trc = new EntityReadCache<VaradhiTopic>(Topic, topicStore.getAllTopics()); `

no subclassing required.

common/src/main/java/com/flipkart/varadhi/common/events/EntityEventListener.java
```diff
@@ -23,5 +23,5 @@ public interface EntityEventListener {
      * @param event the event containing information about the entity change
      * @throws IllegalStateException if the listener is in a state where it cannot process events
      */
-    void onChange(EntityEvent<?> event);
+    void onChange(EntityEvent<T> event);
```
gauravAshok, line 26: due to this change, now the DefaultMetaStoreChangeListener will throw warning.

There we are using it as 
`private final EntityEventListener listener;`
Screenshot 2025-04-14 at 5 10 11 PM


Ad discussed, it probably should be `void onChange(EntityEvent<? extends T> event);`

common/src/main/java/com/flipkart/varadhi/common/events/EntityEventListener.java
```diff
@@ -12,7 +12,7 @@
  * @see EntityEvent
  */
 @FunctionalInterface
-public interface EntityEventListener {
+public interface EntityEventListener<T> {
```
gauravAshok, line 15:
```suggestion
public interface EntityEventListener<T extends MetaStoreEntity> {
```
?

producer/src/main/java/com/flipkart/varadhi/produce/providers/TopicProvider.java
```diff
@@ -0,0 +1,73 @@
...
+ *   <li>Efficient topic lookup by name</li>
+ *   <li>Preloading topics from the metadata store</li>
+ *   <li>Automatic cache updates via entity events</li>
+ * </ul>
+ *
+ * @see EntityProvider
+ * @see VaradhiTopic
+ */
+@Slf4j
+public class TopicProvider extends EntityProvider<VaradhiTopic> {
```
gauravAshok, line 31: as mentioned in previous comment. These classes have less value and we can get away with them if we just use "strategy" pattern / composition.

server/src/main/java/com/flipkart/varadhi/events/EventManager.java
```diff
@@ -0,0 +1,150 @@
...
+ * <p>
+ * This component is designed to be initialized once at application startup,
+ * before any verticles are deployed.
+ *
+ * @see EntityEventDispatcher
+ * @see ProjectProvider
+ * @see TopicProvider
+ */
+@Slf4j
+public final class EventManager {
```
gauravAshok, line 37: Entity word is missing.

server/src/main/java/com/flipkart/varadhi/events/EventManager.java
```diff
@@ -0,0 +1,150 @@
...
+
+    /**
+     * Provider for project entities.
+     */
+    private final ProjectProvider projectProvider;
+
+    /**
+     * Provider for topic entities.
+     */
+    private final TopicProvider topicProvider;
```
gauravAshok, line 47: also instead of having each "readCache" explicitly in here, does it make sense to have it like

Map<ResourceType, ReadCache<?>> so that this class can be used as a singleton to register any readCache once.

server/src/main/java/com/flipkart/varadhi/providers/ProjectProvider.java
```diff
@@ -0,0 +1,73 @@
...
+ *   <li>Efficient project lookup by name</li>
+ *   <li>Preloading projects from the metadata store</li>
+ *   <li>Automatic cache updates via entity events</li>
+ * </ul>
+ *
+ * @see EntityProvider
+ * @see Project
+ */
+@Slf4j
+public class ProjectProvider extends EntityProvider<Project> {
```
gauravAshok, line 31: not required.

gauravAshok (changes requested):
reviewed

common/src/main/java/com/flipkart/varadhi/common/EntityReadCache.java
```diff
@@ -0,0 +1,166 @@
...
+
+        try {
+            log.info("Starting to preload {}", resourceType.toString());
+            List<T> entityList = entityLoader.get();
+
+            if (entityList.isEmpty()) {
+                log.info("No {} found to preload", resourceType);
+                promise.complete();
+                return promise.future();
+            }
```
gauravAshok, line 92: why is this code required?

Below when you created the map, there itself you are printing the size, and in case of 0 size, that will be printed as well.

common/src/main/java/com/flipkart/varadhi/common/EntityReadCache.java
```diff
@@ -0,0 +1,166 @@
...
+     * Returns the entity with the given name.
+     *
+     * @param name the name of the entity to get
+     * @return the entity with the given name, or null if not found
+     * @throws NullPointerException if name is null
+     * @throws ResourceNotFoundException if the entity is not found
+     */
+    public T getEntity(String name) {
+        Objects.requireNonNull(name, "Entity name cannot be null");
+        T entity = entities.get(name);
```
gauravAshok, line 117: just return Optional<T>. no point in throwing the expcetion. The caller can choose to decide what it needs to do with the optional.

common/src/main/java/com/flipkart/varadhi/common/EntityReadCache.java
```diff
@@ -0,0 +1,166 @@
...
+            }
+            case INVALIDATE -> {
+                T removed = entities.remove(entityName);
+                if (removed != null) {
+                    log.debug("Removed {} from local cache: {}", resourceType.toString(), entityName);
+                } else {
+                    log.warn("Attempted to remove non-existent {} from cache: {}", resourceType.toString(), entityName);
+                }
+            }
+            default -> log.warn("Unsupported operation {} for {}: {}", operation, resourceType.toString(), entityName);
```
gauravAshok, line 163: is default really required? try removing it.

common/src/main/java/com/flipkart/varadhi/common/EntityReadCacheRegistry.java
```diff
@@ -0,0 +1,158 @@
...
+     * @param type  the resource type to register the cache for
+     * @param cache the cache instance to register
+     * @return this registry instance for method chaining
+     * @throws NullPointerException if type or cache is null
+     */
+    public <T extends MetaStoreEntity> EntityReadCacheRegistry register(ResourceType type, EntityReadCache<T> cache) {
+        Objects.requireNonNull(type, "Resource type cannot be null");
+        Objects.requireNonNull(cache, "Cache cannot be null");
+
+        caches.put(type, cache);
```
gauravAshok, line 61: check for duplciate registeration.

producer/src/main/java/com/flipkart/varadhi/produce/services/ProducerService.java
```diff
@@ -1,134 +1,267 @@
...
+
+    /**
+     * The region where messages are produced.
+     */
     private final String produceRegion;
 
+    /**
+     * Cache of entity read caches, indexed by resource type.
+     */
+    private final Map<ResourceType, EntityReadCache<?>> readCacheMap;
```
gauravAshok, line 63: why create this hashmap?

We can simply create a field
ReadCache<VaradhiTopic> topicCache;

producer/src/main/java/com/flipkart/varadhi/produce/services/ProducerService.java
```diff
@@ -1,134 +1,267 @@
...
+     * @param cacheRegistry    registry containing entity caches, including the topic cache
+     * @throws NullPointerException if any parameter is null
+     */
     public ProducerService(
         String produceRegion,
-        ProducerOptions producerOptions,
         Function<StorageTopic, Producer> producerProvider,
-        Function<String, VaradhiTopic> topicProvider,
-        MeterRegistry meterRegistry
+        EntityReadCacheRegistry cacheRegistry
```
gauravAshok, line 79: registry should not be a param. ReadCache\<Topic\> should be the param

server/src/main/java/com/flipkart/varadhi/VaradhiApplication.java
```diff
@@ -38,63 +37,122 @@
...
+                        clusterManager,
+                        memberInfo,
+                        cacheRegistry
+                    );
+
+                    // Return both for the next step
+                    return Pair.of(vertx, verticles);
+                })
+            )
+                                                                                     .compose(
```
gauravAshok, line 111: fix this formatting.

server/src/main/java/com/flipkart/varadhi/VaradhiApplication.java
```diff
@@ -104,109 +162,256 @@ private static MemberInfo getMemberInfo(MemberConfig memberConfig) throws Unknow
...
+        )
+                                                                                .setMetricsNaming(
+                                                                                    MetricsNaming.v4Names()
+                                                                                )
+                                                                                .setRegistryName("default")
+                                                                                .addDisabledMetricsCategory(
+                                                                                    MetricsDomain.HTTP_SERVER
+                                                                                )
+                                                                                .setJvmMetricsEnabled(true)
+                                                                                .setEnabled(true);
```
gauravAshok, line 264:
```suggestion
        MicrometerMetricsOptions metricsOptions = new MicrometerMetricsOptions();
        metricsOptions.setFactory(new MicrometerMetricsFactory(meterRegistry))
                      .setMetricsNaming(MetricsNaming.v4Names())
                      .setRegistryName("default")
                      .addDisabledMetricsCategory(MetricsDomain.HTTP_SERVER)
                      .setJvmMetricsEnabled(true)
                      .setEnabled(true);
```

server/src/main/java/com/flipkart/varadhi/VaradhiApplication.java
```diff
@@ -104,109 +162,256 @@ private static MemberInfo getMemberInfo(MemberConfig memberConfig) throws Unknow
...
+            ConfigRetriever retriever = ConfigRetriever.create(vertx, options);
 
-        try {
-            JsonObject content = retriever.getConfig().toCompletionStage().toCompletableFuture().join();
-            AppConfiguration configuration = content.mapTo(AppConfiguration.class);
-            configuration.validate();
-            return configuration;
-        } catch (Exception e) {
-            throw new InvalidConfigException("Failed to load Application Configuration", e);
+            try {
```
gauravAshok, line 347: why split into 2 layer of try catch?

server/src/main/java/com/flipkart/varadhi/cluster/MessageExchange.java
```diff
@@ -54,6 +54,7 @@ public void publish(String routeName, String apiName, ClusterMessage msg) {
     public CompletableFuture<ResponseMessage> request(String routeName, String apiName, ClusterMessage msg) {
         CompletableFuture<ResponseMessage> future = new CompletableFuture<>();
         String apiPath = getPath(routeName, apiName, RouteMethod.REQUEST);
+        log.info("SENDING REQUEST to address: {}", apiPath);
```
gauravAshok, line 57: this log is not required

server/src/main/java/com/flipkart/varadhi/events/EntityEventDispatcher.java
```diff
@@ -0,0 +1,222 @@
...
+     * This set is derived from the listeners map and is immutable.
+     */
+    private final Set<ResourceType> supportedTypes;
+
+    /**
+     * Creates a new EntityEventDispatcher with the specified listeners.
+     *
+     * @param listeners a map of resource types to their corresponding event listeners
+     */
+    private EntityEventDispatcher(Map<ResourceType, EntityEventListener<?>> listeners) {
```
gauravAshok, line 50: why not use registry here?

server/src/main/java/com/flipkart/varadhi/events/EntityEventDispatcher.java
```diff
@@ -0,0 +1,222 @@
...
+            default -> new IllegalArgumentException(errorMsg, cause);
+        };
+
+        if (cause == null) {
+            log.error(errorMsg);
+        } else if (errorMsg.equals(cause.getMessage())) {
+            log.error(errorMsg, cause);
+        } else {
+            log.error("{}: {}", errorMsg, cause.getMessage(), cause);
+        }
```
gauravAshok, line 160: no logging required. I think we are logging too much at many places. This will lead to log pollution only as more and more code is added.

server/src/main/java/com/flipkart/varadhi/events/EntityEventDispatcher.java
```diff
@@ -0,0 +1,222 @@
...
+     *
+     * @param message the cluster message containing the entity event
+     * @param <T>     the type of entity in the event, must extend MetaStoreEntity
+     * @return a future that completes with the response message
+     */
+    @SuppressWarnings ("unchecked")
+    public <T extends MetaStoreEntity> CompletableFuture<ResponseMessage> processEvent(ClusterMessage message) {
+        if (message == null) {
+            return createErrorResponse(null, "Message cannot be null", null);
+        }
```
gauravAshok, line 70: why are we putting defensive null checks everywhere?

There is no reason this method will be called with null message.

server/src/main/java/com/flipkart/varadhi/events/EntityEventDispatcher.java
```diff
@@ -0,0 +1,222 @@
...
+
+        String messageId = message.getId();
+        EntityEvent<?> rawEvent;
+        EntityEvent<T> event;
```
gauravAshok, line 74: "event" is unused

server/src/main/java/com/flipkart/varadhi/events/EntityEventDispatcher.java
```diff
@@ -0,0 +1,222 @@
...
+
+        try {
+            rawEvent = message.getData(EntityEvent.class);
+            if (rawEvent == null) {
```
gauravAshok, line 78: why would this be null?

server/src/main/java/com/flipkart/varadhi/web/v1/admin/DlqHandlers.java
```diff
@@ -36,13 +39,20 @@
 public class DlqHandlers implements RouteProvider {
     private static final long UNSPECIFIED_TS = 0L;
     private final SubscriptionService subscriptionService;
-    private final ProjectService projectService;
+    private final Map<ResourceType, EntityReadCache<?>> readCacheMap;
```
gauravAshok, line 42: just use ReadCache<Project>

gauravAshok (changes requested):
Need some refactoring..

Few rules to follow:
1. Minimize logging.
2. Try to avoid defensive coding. No null checks everywhere. With  null checks it will fail with the same exception. nothing improves and it does not magically handle the null scenarios. It does not avoid bugs. Nullability should be shown as part of documentation / annotations.
3. Defensive null checks are useful in public apis exposed as library / rest layer.
4. Excessive try catches are bad. It reduces readability. If the method can fail, try to show it in its "throws". Method should only catch, if it knows it can handle the failure and recover from it.

server/src/main/java/com/flipkart/varadhi/events/EntityEventDispatcher.java
```diff
@@ -0,0 +1,222 @@
...
+            Object rawResource = rawEvent.resource();
+            T typedResource = null;
+
+            if (rawResource != null) {
+                if (rawResource instanceof Map) {
+                    String json = JsonMapper.jsonSerialize(rawResource);
+                    Class<T> entityClass = getEntityClassForResourceType(resourceType);
+                    typedResource = JsonMapper.jsonDeserialize(json, entityClass);
+                } else if (rawResource instanceof MetaStoreEntity) {
+                    typedResource = (T)rawResource;
```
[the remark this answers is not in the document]
devanshu-garg-IN: It was throwing an exception while testing,

`java.lang.Exception: class java.util.LinkedHashMap cannot be cast to class com.flipkart.varadhi.entities.MetaStoreEntity`

Likely happening during serialization/deserialization when events are sent between nodes in the cluster.

common/src/main/java/com/flipkart/varadhi/common/EntityReadCache.java
devanshu-garg-IN: Changing to `Optional` would require modifying all calling code. All the callers expect this behavior currently and are already set up to handle this pattern of throwing `ResourceNotFoundException` exception.

common/src/main/java/com/flipkart/varadhi/common/EntityReadCache.java
```diff
@@ -0,0 +1,169 @@
...
+     * <p>
+     * This method fetches all entities using the configured entity loader and
+     * populates the cache. It is thread-safe and can be called concurrently.
+     * <p>
+     * The operation is performed asynchronously and returns a {@link Future} that
+     * completes when the preload operation is finished.
+     *
+     * @return a future that completes when the preload operation is finished
+     */
+    public Future<Void> preload() {
```
gauravAshok, line 106: needs to be non public.

common/src/main/java/com/flipkart/varadhi/common/EntityReadCache.java
```diff
@@ -0,0 +1,169 @@
...
+     */
+    @Override
+    public void onChange(EntityEvent<? extends T> event) {
+        String entityName = event.resourceName();
+        EventType operation = event.operation();
+
+        if (operation == EventType.UPSERT) {
+            T entity = event.resource();
+            if (entity != null) {
+                entities.put(entityName, entity);
```
gauravAshok, line 163: we should only put when the version is higher than the current

server/src/main/java/com/flipkart/varadhi/VaradhiApplication.java
```diff
@@ -104,109 +145,238 @@ private static MemberInfo getMemberInfo(MemberConfig memberConfig) throws Unknow
...
+        );
+
+        // Combine futures and register caches when they're ready
+        return Future.all(projectCacheFuture, topicCacheFuture).map(v -> {
+            // Register preloaded caches
+            registry.register(ResourceType.PROJECT, projectCacheFuture.result());
+            registry.register(ResourceType.TOPIC, topicCacheFuture.result());
+            EntityEventDispatcher.bindToClusterEntityEvents(vertx, memberInfo, clusterManager, registry);
+            return registry;
+        });
```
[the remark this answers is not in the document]
gauravAshok: this will be fixed in future.

gauravAshok (approved):
LGTM

producer/src/test/java/com/flipkart/varadhi/services/ProducerServiceTests.java
```diff
@@ -90,58 +120,50 @@ public void testProduceMessage() throws InterruptedException {
...
         Message msg1 = getMessage(0, 1, null, 0);
-        VaradhiTopic vt = getTopic(topic, project, region);
+        String topicName = VaradhiTopic.buildTopicName(project.getName(), topic);
         doReturn(producer).when(producerFactory).newProducer(any());
-        doThrow(new ResourceNotFoundException("Topic doesn't exists.")).when(topicProvider).get(vt.getName());
-        Assertions.assertThrows(
+        when(topicReadCache.get(topicName)).thenThrow(
+            new ResourceNotFoundException(String.format("TOPIC(%s) not found", topicName))
+        );
+        ResourceNotFoundException ex = Assertions.assertThrows(
```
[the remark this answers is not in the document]
gauravAshok: fixed

PR #309, opened 2026-04-15
Update e2e.yml

bandeep18:
## Summary by CodeRabbit

* **New Features**
  * Subscription updates now persist additional subscription metadata (properties and target client identifiers).
  * Queue response payload uses "name" as the queue identifier field.

* **Chores**
  * Improved end-to-end test workflow diagnostics and upgraded test environment components for more reliable troubleshooting.

* **Tests**
  * Updated and strengthened end-to-end and unit tests to reflect API payload/response and default subscription data changes.

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -37,6 +37,11 @@ jobs:
         run: |
           docker compose --profile test -f setup/docker/compose.yml down -v
           docker compose --profile test -f setup/docker/compose.yml up -d --wait --wait-timeout 180
+      - name: Debug Zookeeper on Failure
+        if: failure()  # This only runs if the previous step fails
+        run: |
+          docker compose --profile test -f setup/docker/compose.yml logs zookeeper
+          docker inspect zookeeper
       - name: Run the test
         run: ./gradlew test testE2E
       - name: Publish Test Results
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -35,6 +35,7 @@ jobs:
         run: docker build . --file setup/docker/Dockerfile --tag varadhi.docker.registry/varadhi:latest --build-arg ENV=test
       - name: Setup the environment
         run: |
+          export JMXDISABLE=true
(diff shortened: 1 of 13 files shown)


Mrsandeep27 (commented):
Two unrelated changes bundled here — worth splitting before merge:

**1. CI debug + JMX disable.** Adding \`export JMXDISABLE=true\` and the failure-only zookeeper log dump are fine, but they belong in their own commit / PR titled like \`ci(e2e): disable Zookeeper JMX and add failure diagnostics\`. Mixing CI infra with API-shape changes makes \`git blame\` and \`revert\` harder six months out.

**2. `updateSubscription` adding `properties` and `targetClientIds` parameters.** This is the actual product change. A few things:
- The method signature now takes 8 positional params. That's already over the line where most reviewers (and IDEs) start asking for a parameter object. Consider a \`SubscriptionUpdateRequest\` record with named fields — same wire shape, no risk of \`updateSubscription(... endpoint, retry, consumption, props, clients, requestedBy)\` getting reordered by accident.
- All test mock callsites had to add \`any(), any()\` in the right slot. If a future refactor adds a 9th param, every test breaks again. The parameter object would also fix this.
- \`subscription.setProperties(properties)\` overwrites — if a caller passes \`null\` or an empty map intending \"don't touch\", you'll silently clear existing properties. Worth a guard: only update if non-null, and document the empty-map behavior explicitly.

**3. Test coverage gap.** The new params are exercised via existing tests that just pass \`any()\` to mocks. Worth adding a positive test: \`updateSubscription_withNewProperties_persistsThemToStore()\` that asserts the actual values made it through. Otherwise a future regression where one of the two new params is silently dropped wouldn't be caught.

**4. The CI debug step references `docker compose ... logs zookeeper` and `docker inspect zookeeper`.** If the compose service name isn't \`zookeeper\` (e.g. it's \`zk\` or \`zookeeper-test\`) those commands silently produce empty logs. Worth confirming the service name in \`compose.yml\`.

Direction is right; mostly splitting concerns + a defensive null check.

PR #310, opened 2026-04-15
Version upgrade fix 2

bandeep18:
## Summary by CodeRabbit

* **Chores**
  * Bumped application version to 0.1.8-SNAPSHOT across build and deployment configurations
  * Refactored consumer configuration handling to streamline batch receive policy management

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -2,6 +2,6 @@ projectGroupId=com.flipkart.varadhi
 projectName=varadhi
 majorVersion=0
 minorVersion=1
-patchId=7-SNAPSHOT
+patchId=8-SNAPSHOT
 mavenPublishUrl="" # Add repository URL with double quotes removed
 systemProp.org.gradle.internal.publish.checksums.insecure=true
--- a/pulsar/src/main/java/com/flipkart/varadhi/pulsar/config/ConsumerOptions.java
+++ b/pulsar/src/main/java/com/flipkart/varadhi/pulsar/config/ConsumerOptions.java
@@ -80,23 +80,17 @@ public class ConsumerOptions {
 
     public synchronized Map<String, Object> asMap() {
         Map<String, Object> configMap = new HashMap<>();
-        configMap.put("maxPollRecords", this.maxPollRecords);
-        configMap.put("fetchMaxBytes", this.fetchMaxBytes);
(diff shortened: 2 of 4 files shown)


Mrsandeep27 (commented):
The fix is clear — \`asMap()\` was passing through Pulsar-irrelevant config keys (\`maxPollRecords\`, \`fetchMaxBytes\`, \`fetchMaxWaitMs\`, \`messageSourcePollMaxWaitMs\`, \`maxInMemoryMessages\`) to \`Consumer.loadConf()\` along with a pre-built \`batchReceivePolicy\`, when the right surface for those values is \`ConsumerBuilder.batchReceivePolicy()\`. Switching to the builder API is correct.

A few thoughts:

**1. `buildBatchReceivePolicy()` visibility flip from `private` to `public`.** This now exposes a method that needed to be called from a different package (\`PulsarConsumer\` reaches into \`ConsumerOptions\`). Worth confirming this is the only caller — if it is, package-private (no modifier) keeps the seam tight to the pulsar package without exposing the helper to unrelated consumers of \`ConsumerOptions\`.

**2. The dropped keys aren't accidental dead code.** \`maxPollRecords\`, \`fetchMaxBytes\`, \`fetchMaxWaitMs\` were the input params for \`buildBatchReceivePolicy()\` itself. They're still loaded via the existing config plumbing, just no longer leaked into the loadConf map. Worth a one-line comment in \`asMap()\` saying "fetch tuning fields are now applied via ConsumerBuilder.batchReceivePolicy()" so the next reader doesn't think they were forgotten.

**3. Version bump from `7-SNAPSHOT` to `8-SNAPSHOT`.** Good — but if this PR is going through code-review iteration, the snapshot bump is going to conflict with anyone else who has a draft PR open against the same gradle.properties. Worth merging the version bump as a separate commit (or rebasing) once it's clear this PR is the one landing first.

**4. No test diff.** Is there test coverage that asserts \`Consumer\` is built with the right \`BatchReceivePolicy\`? If \`getPulsarConsumer()\` is the only call site and it's hit by integration tests, that's fine; otherwise, a small unit test that mocks the \`ConsumerBuilder\` and asserts \`.batchReceivePolicy(...)\` was called would prevent regression if someone re-introduces the old map.

Direction is right.

